PR #17, opened 2021-04-22
feat: add vol src

### LINKED ISSUE #12: Return ng-precomputed URLs for MNI template spaces as well
siibra-explorer also reads the MRI-scale data from precomputed URLs. These URLs need to be included in the parcellations definition, and forwarded.

xgui3783:
This PR adds the capability to parse `volumeSrc`, if present, in `space` and/or `parcellation`

It should be backwards compatible with spaces/parc that do not contain these fields.

edit: this PR partially fixes #12


dickscheid (commented):
We should refactor this, there is signficant overlap with the BigBrainVolume class in bigbrain.py. I suggest we change bigbrain.py to cloudvolumes.py and include your logic into it.  In fact, BigBrainVolume might derive from the CloudSource class. To be discussed.

I will merge so the functionality becomes available now.

xgui3783:
I can certainly see some overlaps. 

I will add this issue to discussion.

PR #26, opened 2021-04-28
Enforce region decoding on feature extraction

dickscheid:
RegionFeatures now require a concrete Region object at construction time, not only any region specification like an possibly incomplete region name. As a consequence, FeatureExtractors take an atlas object now in their constructor. This moves  region decoding to the Feature Extraction stage for example for ConnectivityProfile features, make the extraction a bit slower, but downstream feature queries safer.

### DIFF
--- a/examples/BigBrain.ipynb
+++ b/examples/BigBrain.ipynb
@@ -102,7 +102,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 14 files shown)


xgui3783 (approved):
per discussion, I added commits that removed auth token, and added test to prevent regression

Issue #28, opened 2021-04-28
[question] should parcellation introduce `groupBy` functionality


xgui3783:
in siibra-explorer, parcellations are (sometimes) grouped according to, say, functional modes, fiber bundle. Should siibra-python also have this functionality?

If yes to above, how should this be implemented?

- with a parcellation.label attribute, which expects a list of string, so user can filter by label
- with a pseudo path as a string attribute, and the client can reconstruct the tree, to allow for a hierarchy organisation of parcellation
- other suggestions?

pinging @dickscheid @marcenko @skoehnen

dickscheid:
I think that this should be stored as a "modality" attribute of the parcellation (cytoarchitecture, function, fibre architecture, structurual connectivity, functional connectiviy, multimodal)

xgui3783:
perfect! this would actually solve another issue that daviti and I are discussing. 

I can create a quick PR for this?

xgui3783:
I will update the configration repo first. 

To ensure naming consistency, here are the possible values I will add to every `parcellation.modality` (`modalities` ? should this be a list, to be future proof? in the event that some parcellation may have a combination of modalities?) 

- cytoarchitecture (julich brain v1.18, julich brain v2.5)
- fibre architecture (long bundle, short bundle, hcp bundle)
- functional modes (difumo [] )

these do not  yet have modalities assigned:
- cortical layers
- isocortex
- dk

xgui3783:
fixed by #32

PR #32, opened 2021-04-29
feat: add modality and version parsing

### LINKED ISSUE #29: parcellation missing `.version` metadata
for example, julich brain v2.5 is the next version of julich brain v1.18. 

This is, I believe, currently missing. 

Together with versioning metadata, we should also introduce `shortname` for the version (not just the parcellation)

xgui3783:
fixes #29 and #28


xgui3783:
also related is the PR that update config file: https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/3

PR #39, opened 2021-05-11
feat: allow kg datasets to be fetched

### LINKED ISSUE #21: Implement access to all linked EBRAINS features in python client

xgui3783:
fixes #21


dickscheid (commented):
I would modify the "kg" abbreviation to "ebrains" in a future patch, it is better interpretable for users.

PR #46, opened 2021-05-18
Ebrains regional datasets accessible in example notebooks

dickscheid:
I polished the EBRAINS feature query that was recently implemented into the development branch. I  renamed the symbols with „Kg“ that are visible to users to „Ebrains“, because it is better interpretable.

One thing that I realized is that the EBRAINS summary query gives a  lot of results that are not datasets, but sample, like this one: 'consc_j12.prop1-r1  ' (id:https://nexus.humanbrainproject.org/v0/data/minds/experiment/sample/v1.0.0/b77a0bb8-5fd9-4e78-a14e-92bb1408c040)

While there might be interesting search matches in there, the full query for the details will fail on those since it only interprets datasets, not experiments and samples. So this branch filters them already after the summary query.

It might be more efficient to modify the EBRAINS query instead, so they wouldn’t even be requested from the KG. 

It would also be good to have a few more attributes in the summary to be able to filter the datasets before fetching the details from the KG. We should discuss this...

I moved the query spec out of the python file into a json. While we could, we do not even need to include that json in the release package since it is only used to upload the query once, and not any more when a user works with siibra.


siibra/features/ebrainsquery.py
```diff
@@ -0,0 +1,123 @@
...
+def set_specs():
+  MODULE_DIR=os.path.dirname(os.path.abspath(__file__))
+  with open(os.path.join(MODULE_DIR,'ebrainsquery_specs.json'),'r') as f:
+    QUERYSPEC=json.load(f)
+  req = ebrains.upload_schema(
+    query_id=KG_REGIONAL_FEATURE_SUMMARY_QUERY_NAME,
+    spec=QUERYSPEC['summary_spec'],
+    **kg_feature_summary_kwargs)
+  if not req.status_code < 400:
+    raise RuntimeError("Could not upload query")
```
xgui3783, line 120: tab space == 4?

PR #47, opened 2021-05-20
feat: add support for additional maptypes

xgui3783:
This PR adds support for additional maptypes (possibly needed for julich brain v2.7.3)

additionally, if maptypes is not in a predfined list, logger will warn, rather than throwing.


dickscheid:
@xgui3783 I do not hitnk we need MapType.DISCRETE, this is what MapType.LABELLED was thought for. Would you agree?

xgui3783:
I agree.

This PR was created before or discussion yesterday, and is no longer relevant.

PR #63, opened 2021-06-28
feat: add origin data info

xgui3783:
This PR adds the ability for siibra-python to parse, fetch and make available metadata related to spaces, parcellation and regions. (e.g. id added per https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/11/diffs?commit_id=7aa9a29a445f0d8678c47a562f2dcea124367bae )

For now, there exist two types of `origin_datainfo`, simple or ebrains. 

- simple datainfo -> name, property, urls attribute are parsed as json
- ebrains datainfo -> kgId is parsed, and the above information is fetched from KG.

### DIFF
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -15,6 +15,7 @@
 import re
 from enum import Enum
 from abc import ABC, abstractmethod
+from . import logger
 
 class Glossary:
     """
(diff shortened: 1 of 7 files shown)


xgui3783:
This PR should be backwards compatible with previous version(s) of configuration.

marcenko:
I will take a look and see if I can merge it.

xgui3783:
@marcenko do you think it make sense to extend connectivity profile (and matrix) to use origin data attribute? (from https://github.com/FZJ-INM1-BDA/siibra-python/pull/59)

In essence, they are describing something very similar.

I can provide a PR if it's suitable?

marcenko:
@xgui3783 I could imagine that it fits. At least as far as I understood the origin data.

PR #61, opened 2021-06-23
feat: improve the response time of (ebrains) features

xgui3783:
This PR does two things to improve the loading time of feature extractor(s):

- reuse feature extractor (currently, each call to `atlas.get_features` results the instantiation of a new feature extractor)
- (for `EbrainsRegionalFeature` implements static method `get_cache`, which can be called ahead of time to further improve response time)

This should result in drastic improvement in:

- calling `get_feature`, especially:
  - repeated calling `get_feature` on same modality
  - if modality is ebrainsfeatures
  - if modality is ebrainsfeatures, and `get_cache` is called prior to calling `get_feature`

I will paste some performance comparison in the comments

---

**nb**

this PR also changes the way users should instantiate feature extractors.

Instead of passing the entire atlas object, one must explicitly pass parcellation and region as positional arguments, if required. The feature extractor will raise an exception if a positional argument is required, but not supplied.

This should allow caching of extractors to function properly.

PR #64, opened 2021-06-30
feat: improve the response time of (ebrains) features

xgui3783:
supercedes #61

After discussion, this is a simplified, but equally effective:

```sh
--- preheat-1: 242.01770949363708 seconds ---
--- preheat0: 242.37783551216125 seconds ---
--- preheat1: 0.36009788513183594 seconds ---
--- with preheat again: 0.3659327030181885 seconds ---
```

### DIFF
--- a/siibra/atlas.py
+++ b/siibra/atlas.py
@@ -349,7 +349,7 @@ class Atlas:
                 extractor = cls(self)
             hits.extend(extractor.pick_selection(self))
 
-        return hits
+        return list(set(hits))
 
     def assign_coordinates(self,space:Space,xyz_mm,sigma_mm=3):
         """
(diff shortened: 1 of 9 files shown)


xgui3783:
@dickscheid @marcenko can you please take a look at the rework of the PR

The performance is not as quick (0.4sec vs 0.01sec), but already a great improvement. The benefit is that change of repo is minimal, and should be more readable.

xgui3783:
Marked as draft for now. Found a bug. Well provide a fix, then mark as ready.

xgui3783:
Marked as draft for now. Found a bug, will provide a fix today. Will mark as ready when done.

xgui3783:
The failing test is (indirectly) due to the errornous implementation of the connectivity profile (and potentially connectivity matrix)

I will raise a separate PR to fix that issue. Only when the tests passes, I will open this PR for review.

PR #66, opened 2021-07-01
bugfix: error parsing of conn profile

xgui3783:
This PR fixes an issue discovered by https://github.com/FZJ-INM1-BDA/siibra-python/pull/64#issuecomment-872186814

Up until this point, the parsing of connectivity feature is faulty:

- unparsable region names are discarded
- corresponding profile entry is not trimmed
- during parsing, regionnames and profile are zipped via list comprehension, which results in misalign of profile value -> region

This one of the reasons for the IndexError https://github.com/FZJ-INM1-BDA/siibra-python/blob/8d0a632/test/features/test_connectivity.py#L38 and https://github.com/FZJ-INM1-BDA/siibra-python/blob/8d0a632/test/features/test_connectivity.py#L43

This PR is... potentially less backwards compatible. column_names is no longer a list, but a dictionary, with int as keys. ConnectivityProfile __str__ has an example (https://github.com/FZJ-INM1-BDA/siibra-python/blob/3b99a54/siibra/features/connectivity.py#L61) of how to handle  when a region was not able to be decoded properly

### DIFF
--- a/siibra/features/connectivity.py
+++ b/siibra/features/connectivity.py
@@ -58,7 +58,7 @@ class ConnectivityProfile(RegionalFeature):
             vrange = self.globalrange
         calib = termplot.calibrate(vrange, self.column_names)
         return "\n".join([style.BOLD+'Connectivity profile of "{}"'.format(self.region)+style.END] 
-                + [ termplot.format_row(self.column_names[i],profile[i],calib)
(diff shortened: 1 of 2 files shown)


xgui3783:
@dickscheid and @marcenko can you take a look that this PR?

particularly @marcenko if this might break anything in siibra-api (connectivity_profile.column_names is no longer a list, but a dictionary with int as accessors)

PR #70, opened 2021-07-13
bugfix: conn profile when parc switches ...

xgui3783:
...cache does not clear
chore: fix unit tests


xgui3783:
@dickscheid can you take a look at this PR. 

on trying to fix the unit test, I discovered another bug re connectivity profile extractor.

Prior, the extractor will fetch datafeature, parse them, filter them (according to selected parcellation !! ) and store as a class property once and only once on init of extractor.

This caused the undesirable behaviour where if the user first tries to fetch connectivity profile for julich brain 1.18, then switches parcellation to julich brain 2.9.0, the connectivity profile features will always return as an empty array 

This PR should fix that (as tested via https://github.com/FZJ-INM1-BDA/siibra-python/blob/pr_connProfileExtr/test/features/test_connectivity.py#L5-L20)

dickscheid:
This is a good approach. I am suprprised that this problem did not reach the surface earlier.

xgui3783:
Believe it or not, the failed tests is actually another bug. I am in the process of fixing it at the moment.

PR #72, opened 2021-07-19
bugfix: fixing deprecated flag parse

xgui3783:
fixes how deprecated flag is parsed


xgui3783:
If you can wait for a few minutes, I would like to add tests to prevent regression

PR #73, opened 2021-07-19
Spatial region properties now fully precomputed in the brainscapes-da…

dickscheid:
…tafeatures repository.


xgui3783:
this one looks good to me. If we merge to develop, I can actually work on siibra-api integration rightaway!

xgui3783:
(and maybe then, I will find some bugs =S)

PR #74, opened 2021-07-19
Local caching of siibra configurations

dickscheid:
This PR implements basic caching of configurations. The cache is sensitive to the gitlab config site and project tag. Cached files are written to default CACHEDIR defined in the retrieval module.

### DIFF
--- a/siibra/config.py
+++ b/siibra/config.py
@@ -16,12 +16,16 @@ import json
 from . import logger,__version__
 from .commons import create_key
 from .retrieval import CACHEDIR
-from gitlab import Gitlab, exceptions as gitlab_exceptions
+from gitlab import Gitlab
+from gitlab.exceptions import GitlabError
+from requests.exceptions import ConnectionError
 from tempfile import mkstemp
 from tqdm import tqdm
 import os
 from collections import defaultdict
 import re
+from datetime import datetime
+from glob import glob
 
 # Until openminds is fully supported, 
 # we store atlas configurations in a gitlab repo.
@@ -39,7 +43,7 @@ class ConfigurationRegistry:
     repository (atlases, parcellations, spaces).    
     """
 
-    GITLAB_CONFIGS=[{
+    GITLAB_CONFIGURATION_REPOSITORIES=[{
         'SERVER': 'https://jugit.fz-juelich.de',
         'PROJECT_ID': 3484,
     }, {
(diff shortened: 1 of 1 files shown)


dickscheid:
I just realize that this implementation has one problem: Configurations are identified by the rpoject tag, e.g. "develop". If this is not a git tag, but a branch, siibra will not realize that the configuration behind this tag has changed and continue to deliver the cached version until it is deleted. We should probably add a detection mechanism for the project tag, so that caching is only applied for real tags / commits, not for branch names. @xgui3783 any idea how to do that?

dickscheid:
Solved. The module only applies caching now if the project tag in gitlab is really a tag. This should do. Tested it for SIIBRA_CONFIG_GITLAB_PROJECT_TAG='develop' (not cached) and SIIBRA_CONFIG_GITLAB_PROJECT_TAG='siibra-0.1a6' (cached).

siibra/config.py
```diff
@@ -68,6 +77,8 @@ def __init__(self,config_subfolder,cls):
                 logger.debug(f'Attempting to connect to {GITLAB_SERVER}')
                 # 10 second timeout
                 project=Gitlab(gitlab_config['SERVER'], timeout=10).projects.get(GITLAB_PROJECT_ID)
+                if GITLAB_PROJECT_TAG in map(lambda t:t.name,project.tags.list()):
+                    activate_caching = True
```
xgui3783, line 81: I understand this is to check if `GITLAB_PROJECT_TAG` is a tag or branch. 

I had the idea of using `config_{GITLAB_PROJECT_TAG}_{COMMIT_HASH}/{directory}_{fname}.json`

rather than `config_{GITLAB_PROJECT_TAG}_{config_folder}_{fname}.json`

so we could simply check if the directory `config_{GITLAB_PROJECT_TAG}_{COMMIT_HASH}` exists.

this way, we **could** cache branches, too, and would speed up dev process (for siibra-api)

siibra/config.py
```diff
@@ -76,33 +87,62 @@ def __init__(self,config_subfolder,cls):
...
+                f.close()
+
+        # activate cache only if the gitlab project tag was a protected tag. 
+        # For other tags (e.g. if the project tag is a branch), 
+        # the cache will be updated each time siibra is loaded.
+        if activate_caching:
+            cachefile = os.path.join(CACHEDIR,f"config_{GITLAB_PROJECT_TAG}_{config_folder}.json")
+            with open(cachefile,'w') as f:
+                json.dump(config,f,indent='\t')
+
```
xgui3783, line 121: perhaps we should add a vanity file to indicate cache store of all relevant files are completed.

otherwise, if user keyboard interrupts at the wrong time, we could have an incomplete cache, which could be a mess to debug.

siibra/config.py
dickscheid: Or would you suggest to always connect to the server to get info on the tag? This would mean you cannot work offline.

siibra/config.py
dickscheid: This one can actually be considered a vanity file, no? It is the last file written after the actual configuration jsons are already there. If this file is not written, the cache is not used and recomputed next time.  I would think that if a user interrupts during caching, this last file will usually not be created.

siibra/config.py
dickscheid: The current solutions does not connect to servers at all if a cache is found, which is a big step forward since it allows basic offline use of siibra after one successful run with a given configuration tag.

siibra/config.py
dickscheid: > aww, I think I misunderstood.
> 
> In this scenario, an entire folder (e.g. an entire space registry) is saved as a single json file.
> 
> this is probably not needed then.

It is not the complete registry, but the paths to all locally stored files for the registry. So effectively, you can say it captures the complete registry once it is retrieved.

siibra/config.py
xgui3783: I see. 

am I understanding correctly that the the improvement is:

| | number of network requests in second import |
| --- | --- |
| v0.1a6 | 30+ |
| this branch, for tags | 0 |
| this branch, for branch (changed, or unchanged) | 30+ |
| my proposal, for tags | 1 |
| my proposal, for branches, unchanged | 1 |
| my proposal, for branches, changed | 30+ |

xgui3783: in fact, we can have a hybrid approach:

if `GITLAB_PROJECT_TAG` starts with `siibra-`, we can assume it is a tag, always try to look for, and cache it like it's a tag.

otherwise, always treat it as if it is a branch, and compare hash.

In siibra-configuration, we need to make this very clear, how to release new tags.

dickscheid: Yes I think this is correct. A tradeoff to be made. Woul it be a solution to use dev-tags for development? that would work as well.

dickscheid: We could also implement your proposal, catch the offline case, and fall back to last cached version of the branch if the gitlab server cannot be reached (with an info message). What do you think?

dickscheid: > in fact, we can have a hybrid approach:
> 
> if `GITLAB_PROJECT_TAG` starts with `siibra-`, we can assume it is a tag, always try to look for, and cache it like it's a tag.
> 
> otherwise, always treat it as if it is a branch, and compare hash.
> 
> In siibra-configuration, we need to make this very clear, how to release new tags.

I am not a big friend of interpreting strings for such cases, since it assumes extra knowledge and discipline, even if we document it somewhere. 

What about my proposal: test wether system has internet access, if yes: clarify tag/commit-hash, otherwise: take most recent cached version of the given tag and output info message.

dickscheid: are there good ways of checking internet access?

xgui3783: so an example flow would be:

if ref starts with `siibra-`:

- check if cache available
    - if yes, use cache
    - if no, get from source, and store as cache

otherwise:

- fetch hash of git ref
    - if fails, check if cache exists
        - if cache exists, use cache
        - if cache does not exist, raise Exception
    - if succeeds, check if cache exists
        - if cache exists, compare hash
            - if same hash, use cache
            - if not same hash, fetch all files, store them in cache
        - if cache does not exist, fetch all files, store them in cache

dickscheid: I will give that a try this morning!

xgui3783: sorry, I didn't see your interim messages

I agree with your sentiment. checking string is fragile.

I actually like your idea a lot more (clarify commit hash, if cannot, check cache)

xgui3783: and the logic would be **a lot** simpler

dickscheid: good, I'll go for it

xgui3783: re: check if internet connection is available, if unreachable, `Gitlab` would simply raise http exception

we should be mindful of instances where mirrors are offsync. Ideally we check cached commit hash against remote commit hash, to see if remote is indeed newer.

the more we talk about this, the more I feel this is more complicated than it needs to be. Maybe we are overengineering a simple scenario.

perhaps, a simpler approach may be:

- cache any ref, logger.info the commit hash
- if developer would like to refresh the cache, they can simply clear the cache folder

xgui3783: the more we talk about this, the more I feel this is more complicated than it needs to be. Maybe we are overengineering a simple scenario.

perhaps, a simpler approach may be:

- cache any ref, logger.info the commit hash
- if developer would like to refresh the cache, they can simply clear the cache folder

dickscheid:
@xgui3783 could you have a look at the latest commit? It implements our ideas above, in particular:

- if config repository reached, test wether configuration project tag is a branch or not
- if requested configuration project tag is a branch, perform caching with concrete commit id / datestamp
- if config repository cannot be reached, take what is available - for branches, the most recent datestamp; for real tags all fine anyway

I tested most cases, except for new commits to a branch, since I didn't want to touch configurations only for testing.

xgui3783:
looks good to me!

xgui3783:
@dickscheid let me know if you want me to merge, and when you release a new version of siibra-python?

Issue #101, opened 2021-10-13
Add jsonable abstract class or decorators for easy json encoding


xgui3783:
To consider also: avoid circular references

xgui3783:
from the discussion with @dickscheid , we have made the following way forward:

- a prototype refactor of siibra-python will be attempted, on https://github.com/FZJ-INM1-BDA/siibra-python/tree/refactor_rigourous_basemodels , in which:
  - [ ] classes will directly sublcass `BaseModel`
  - [ ] classes will follow schema from openminds
  - [ ] the experiment will start small, with the following classses:
    - [ ] Location
    - [ ] Point
    - [ ] BoudningBox
    - [ ] Space
- Potential issues:
  -  [ ] compatibility with openminds (e.g. key such as `space_id` vs `coordinateSpace`)
  -  [ ] possible to keep the current data structure (e.g. volume.space reference the space instance, useful in python) and openminds compliant (`coordinateSpace` complies to openminds schema)

xgui3783:
according to https://github.com/samuelcolvin/pydantic/issues/643 , a 3rd party library https://github.com/koxudaxi/datamodel-code-generator exists to generate pydantic models directly from json schema

PR #105, opened 2021-10-22
feat: first prototype of jsonable abstract cls

xgui3783:
First pass at fixing #101 

What has been done:

- Abstract class implemented (https://github.com/FZJ-INM1-BDA/siibra-python/blob/feat_jsonableEncoders/siibra/core/concept.py#L265-L279)
- Extraction mechanism implemented (https://github.com/FZJ-INM1-BDA/siibra-python/blob/feat_jsonableEncoders/siibra/core/json_encoder.py)
- Tests for extraction mechanism implemented (https://github.com/FZJ-INM1-BDA/siibra-python/blob/feat_jsonableEncoders/test/core/test_json_encoder.py)
- A number of simple implementation:
  - parcellation: (https://github.com/FZJ-INM1-BDA/siibra-python/blob/d529fa9b2d28aa97c79c05488ac8188cfd8069bf/siibra/core/parcellation.py#L444-L459)
  - region (https://github.com/FZJ-INM1-BDA/siibra-python/blob/d529fa9b2d28aa97c79c05488ac8188cfd8069bf/siibra/core/region.py#L820-L842)
  - space (https://github.com/FZJ-INM1-BDA/siibra-python/blob/d529fa9b2d28aa97c79c05488ac8188cfd8069bf/siibra/core/space.py#L135-L146)
  - point (https://github.com/FZJ-INM1-BDA/siibra-python/blob/d529fa9b2d28aa97c79c05488ac8188cfd8069bf/siibra/core/space.py#L533-L546) (bonus point for being sands comforting)
- example notebook (https://github.com/FZJ-INM1-BDA/siibra-python/blob/d529fa9b2d28aa97c79c05488ac8188cfd8069bf/examples/JSONables.ipynb)


siibra/core/parcellation.py
```diff
@@ -445,3 +449,37 @@ def _from_json(cls, obj):
...
+        name: Optional[str]
+        # some parcellation have None as modality
+        modality: Optional[str]
+        # turns out, volumesrc is a sublcass of dataset
+        infos: Optional[List[Union[VolumeSrc.SiibraSerializationSchema, Dataset.SiibraSerializationSchema]]]
+        regions: Optional[List[Region.SiibraSerializationSchema]]
+        spaces: Optional[List[Space.SiibraSerializationSchema]]
+        volumes: Optional[List[VolumeBaseModel]]
+
+    SiibraSerializationSchema = ParcellationSerializationSchema
```
dickscheid, line 465: @xgui3783 I find this pattern very hard to understand (defining a new derived class inside another class definition, then setting a member to this new class ). Couldn't we more obviously implement or overwrite a simple property of the base class SiibraSerializable here? If not, thiis should definitly be documented so to give an idea what's happening...

siibra/core/jsonable.py
```diff
@@ -0,0 +1,39 @@
...
+    from typing import Literal
+
+class SiibraBaseSerialization(BaseModel):
+    at_id: str = Field(alias='@id')
+    at_type: str = Field(alias='@type')
+
+    class Config:
+        arbitrary_types_allowed = True
+
+class SiibraSerializable(ABC):
```
dickscheid, line 19: Shouldn't we simply derive SiibraSerializable from BaseModel? Why defining separate models instead of taking our classes (Atlas, etc.) right away as the models? I think this would make the pydantic integration easier to read and maintain. I understand it might have more implications now, but I find it quite confusing to define all these separate XxxModel classes in the derived classes, and (see other comment) do not like that we import the pydantic BaseModel at so many places in addition to SiibraSerializable.

siibra/core/parcellation.py
```diff
@@ -13,15 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .space import Space
+from .datasets import Dataset
+from siibra.core.json_encoder import JSONEncoder
```
dickscheid, line 17: wouldn't that just be `from .json_encode import ...`?

siibra/core/parcellation.py
```diff
@@ -13,15 +13,18 @@
...
+from .space import Space, VolumeBaseModel
```
dickscheid, line 18: Here we have a good example: If SiibraSerializable were derived from BaseModel, and VolumeSrc a SiibraSerializable, then we need not have this extra class here

siibra/core/jsonable.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @property
+    @abstractproperty
+    def SiibraSerializationSchema(self) -> Union[BaseModel, Union[BaseModel, BaseModel]]:
+        pass
+
+    @abstractmethod
+    def to_json(self, detail=False, **kwargs):
```
dickscheid, line 27: isn't pydantic's model.json() builtin supposed to provide this already, assuming the model/class has been designed appropriately? After reading through the pydantic concepts, I have a feeling that we are here somewhat duplicating its functionality.

https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson

xgui3783:
@dickscheid 

I think you removed a comment, somehow?

I was taking a look at a alternatives, how do you think of the following:

```python
# in parcellation.py

from pydantic import BaseModel

class ParcellationSerializationSchema(BaseModel):
  id: str
  name: str
  # other attributes

class Parcellation(
    AtlasConcept,
    SiibraSerializable,
):
  # ...
  def to_json(self) -> ParcellationSerializationSchema:
    # complex logic
    return {
       'id': 'foo-bar',
       'name': 'hello world'
    }

```

dickscheid (commented):
I find that we introduce here quite many changes throughout the code, given that the serialization is a well defined central concept, with strong support from pydantic. I hope that we can incoporate pydantic more rigorously. See especially comments in the jsonable module. Woudl be good to discuss this.

siibra/core/region.py
```diff
@@ -13,8 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from pydantic import BaseModel
```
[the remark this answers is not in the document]
xgui3783: this import is not used. so I will remove this.

siibra/core/region.py
```diff
@@ -46,7 +51,20 @@
...
+    name: str
+    children: List['RegionSerializationSchema']
+    rgb: Optional[List[int]]
+    centroids: Optional[List[Point.SiibraSerializationSchema]]
+    infos: Optional[List[Dataset.SiibraSerializationSchema]]
+    label: Optional[int]
+    volumes: Optional[List[VolumeBaseModel]]
+    has_regional_map: Optional[bool] = Field(alias='hasRegionalMap')
+
+RegionSerializationSchema.update_forward_refs()
```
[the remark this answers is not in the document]
xgui3783: Self referencing. I will add a comment

siibra/core/region.py
```diff
@@ -46,7 +51,20 @@
 
 REGEX_TYPE=type(re.compile('test'))
 
-class Region(anytree.NodeMixin, AtlasConcept):
+
+class RegionSerializationSchema(SiibraBaseSerialization):
```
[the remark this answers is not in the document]
xgui3783: In essense,  I needed to call `RegionSerializationSchema.update_forward_refs()`

dickscheid:
> @dickscheid
> 
> I think you removed a comment, somehow?
> 
> I was taking a look at a alternatives, how do you think of the following:
> 

See my review. I would be motivated to be more rigorous, getting rid of the separate models and making our core classes the models. When reading the pydantic documentation, my feeling is that this is the intended way. I admit it might cause more work now to review definitions of class members, but my feeling is that it is the cleaner solution. Then, if done right, I also think that pydantic's model.json() function will provide us the explicit to_json() construct then - with some optimism that the circular dependencies can be handeld in a clever way.

dickscheid:
I just saw this: https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude
Maybe along these lines, it will become possible to introduce what I suggest above incrementally?

siibra/core/jsonable.py
xgui3783: My understanding is, there are likely cataclysmic changes to siibra-python, and will require a rewrite if we subclass `SiibraSerializable` from `BaseModel`.

there are some very opinionated how an instance is (and is not) instantiated.

Rather than be saddled with implementaiton details. I figured this is the fastest way to implement a compromise.

siibra/core/parcellation.py
xgui3783: yes, I will fix (my IDE wants to be clever sometimes)

siibra/core/parcellation.py
xgui3783: We could have used `VolumeSrc.SiibraSerializationSchema` instead of `VolumeBaseModel` in this case. (In fact, I should have)

siibra/core/jsonable.py
xgui3783: I think `pydantic` leverages the `dict` built in.

But I want it to be more flexible. 

For example, `Region`'s `to_json` method allow for a `space` kwarg, which, if supplied, will return the centroid.

xgui3783:
> I just saw this: https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude Maybe along these lines, it will become possible to introduce what I suggest above incrementally?

Please see https://github.com/FZJ-INM1-BDA/siibra-python/pull/105#discussion_r748768121 and https://github.com/FZJ-INM1-BDA/siibra-python/pull/105#discussion_r748768749

I think not only would the attempt for `siiibraserizable` to subclass `basemodel` require a substantial rewrite, we will have to patch `basemodel`'s `to_json` call, when the situation demands.

xgui3783:
Ideally, I would vote in favour of move all classes in siibra-python to subclass `BaseModel`.

But I want to believe that the current model is a nice compromise, especially we might need a rewrite when we move from gitlab repo to openminds.

dickscheid:
> > I just saw this: https://pydantic-docs.helpmanual.io/usage/exporting_models/#advanced-include-and-exclude Maybe along these lines, it will become possible to introduce what I suggest above incrementally?
> 
> Please see [#105 (comment)](https://github.com/FZJ-INM1-BDA/siibra-python/pull/105#discussion_r748768121) and [#105 (comment)](https://github.com/FZJ-INM1-BDA/siibra-python/pull/105#discussion_r748768749)
> 
> I think not only would the attempt for `siiibraserizable` to subclass `basemodel` require a substantial rewrite, we will have to patch `basemodel`'s `to_json` call, when the situation demands.

It seems to me that the model.json() can be quite flexibly influenced using pydantic's Config, where json encoder functions could be stored. I do not yet understand all it can do though.

dickscheid:
>  especially we might need a rewrite when we move from gitlab repo to openminds.

That is true. In fact, we can start to move to openminds now. We should discuss the PR in this context the next days.

marcenko:
@xgui3783 Is this independent from the pull request to "Refactor/pydantic". 
I was kind of thinking they are related and have therefore not yet merged it

xgui3783:
> @xgui3783 Is this independent from the pull request to "Refactor/pydantic". I was kind of thinking they are related and have therefore not yet merged it

correct. This PR will be superceded by #143

PR #138, opened 2021-11-18
Feature/surface access

dickscheid:
This PR adds explicit basic surface support:

- new module volumes.mesh
- classes for Gifti meshes and Gifti labels (neuroglancer meshes not yet completely supported)
- clearer distinction of volumes and image volumes (meshes are also volumes, but not image volumes)

I have chosen to return meshes as a dict of two arrays - 'verts' and 'faces'.  The mesh "fetch()" method allows to specify variants, in order to select e.g. white matter, pial, or inflated.

An Example:

```python
import siibra
atlas = siibra.atlases['human']
julich_cmap = atlas.get_parcellation().get_colormap()
julich_surf = atlas.get_map(space="fsaverage").fetch_all(variant='inflated')
# yields a dict with three arrays: verts, faces, labels

from nilearn import plotting
plotting.view_surf(
    surf_mesh=[julich_surf['verts'], julich_surf['faces']], 
    surf_map=julich_surf['labels'],
    cmap=julich_cmap, symmetric_cmap=False, colorbar=False)
```

### DIFF
--- a/siibra/features/feature.py
+++ b/siibra/features/feature.py
@@ -181,7 +181,7 @@ class RegionalFeature(Feature):
     TODO store region as an object that has a link to the parcellation
     """
 
-    def __init__(self, regionspec: Tuple[str, Region], species = [], **kwargs):
+    def __init__(self, regionspec: Tuple[str, Region], species: dict, **kwargs):
         """
         Parameters
         ----------
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -29,18 +29,22 @@ from tqdm import tqdm
 from abc import abstractmethod
 from typing import Union
 
-# Which types of available volumes should be preferred if multiple choices are available?
-PREFERRED_VOLUMETYPES = ["nii", "neuroglancer/precomputed", "detailed maps"]
-
 
 def create_map(parcellation, space: Space, maptype: MapType):
     """
     Creates a new ParcellationMap object of the given type.
     """
-    classes = {
-        MapType.LABELLED: LabelledParcellationMap,
-        MapType.CONTINUOUS: ContinuousParcellationMap,
-    }
+    if space.type=="gii":
+        classes = {
+            MapType.LABELLED: LabelledSurface,
+            MapType.CONTINUOUS: None,
+        }
+    else:
+        classes = {
+            MapType.LABELLED: LabelledParcellationVolume,
+            MapType.CONTINUOUS: ContinuousParcellationVolume,
+        }
+
     if maptype in classes:
         obj = classes[maptype](parcellation, space)
     elif maptype is None:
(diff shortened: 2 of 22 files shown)


siibra/features/feature.py
```diff
@@ -181,7 +181,7 @@ class RegionalFeature(Feature):
     TODO store region as an object that has a link to the parcellation
     """
 
-    def __init__(self, regionspec: Tuple[str, Region], species = [], **kwargs):
+    def __init__(self, regionspec: Tuple[str, Region], species: dict, **kwargs):
```
xgui3783, line 184: should we strictly define each regional feature to be of only a single species?

From KG, this field will always return as a list (KG is not asserting an opinion, it's just how it always return linked entities). 

If we expect a single species for regional dataset, we will need to consider how to deal with regional datasets that **may** have multiple species linked

siibra/volumes/parcellationmap.py
```diff
@@ -909,4 +946,85 @@ def colorize(self, values: dict):
                         affine = thismap.affine
                     result = np.maximum(result, thismap.get_fdata()*value)
                     
-            return Nifti1Image(result, affine) 
\ No newline at end of file
+            return Nifti1Image(result, affine)
+
+
+class LabelledSurface(ParcellationMap):
```
xgui3783, line 952: would it be more appropriate to call this `GiiLabelledSurface`, owing to the assertion in the `__init__` calls?

probably also related, when/if neuroglancer mesh does become supported, would it be another class `NgLabelledSurface(ParcellationMap)` or do you foresee they share the same class?

xgui3783 (commented):
We should discuss how we would like `RegionalFeature` to be structured (per comment)

Otherwise, I cannot immediately see any obvious issues with this first implementation of support for surface.

marcenko:
Seems fine to me. 
Unfortunately I can't say much to the multiple species part.

One test is failing. I did not yet look deeper into it but I can check it tomorrow.

siibra/volumes/parcellationmap.py
dickscheid: they would be two classes, but we would probably subclass this one then. I would prefer to go with the general name for now.

xgui3783:
I investigated and it appears that the issue arise from https://github.com/FZJ-INM1-BDA/siibra-python/blob/2342316/siibra/core/region.py#L153 

for whatever reason, julich brain 1.18 had several `CM (Amygdala) - left hemisphere`'s scattered about, and each has different index. 

this causes `parcellation.decode_region` to return multiple regions when Connectivity entries are being populated.

marcenko:
@dickscheid @xgui3783 
We said in the dev meeting, that this is no longer needed? Just making sure before I remove it.

xgui3783:
I'm quite positive this PR and branch is no longer needed

xgui3783:
per meeting on 2021 01 11, this PR is superceded by commits made directly on main.

PR #142, opened 2021-12-01
bugfix: ieeg feature query space filtering & more predictable region.defined_in_space behaviour

xgui3783:
this PR fixes two issues:
- provides some predictability to `Region.has_regional_map` method (prior, it may raise `RuntimeError` if space provided has no suitable maps)
- fixes ieeg session query based on provided `space` kwarg 

@ reviewers, it could be argued that all spatial queries should adopt the proposed change. Should I amend the PR as such?


dickscheid:
We will reintroduce iEEG from scratch in a version following v0.4. Suggest to delete this PR - ok @xgui3783 ?

PR #145, opened 2021-12-07
feat: add docker image build in CI

xgui3783:
first prorotype of building docker image as a part of CI

It is currently set that on release new docker image will be built and pushed to ebrains docker registry.


skoehnen:
I like the way you prevent the apt-cache from increasing the image size :+1:

xgui3783:
> @xgui3783 I like the way you prevent the apt-cache from increasing the image size +1

still 2GB uncompressed. =S 

need to dig into dep (hopefully most is cloudvolume, but it seems numpy is pretty heavy)

xgui3783:
also nice to have is to have a separate step that tests the built docker image, running in windows, ubuntu and macos

marcenko:
I tried using:
[jupyter/minimal-notebook](https://hub.docker.com/r/jupyter/minimal-notebook)
instead of
[jupyter/scipy-notebook](https://hub.docker.com/r/jupyter/scipy-notebook)

On my local machine the image size decreases from 4.3GB to 2.6GB.

Do we need the bigger one?

xgui3783:
Honestly, I don't know. If all the functionalities are there, we could
definitely switch.

marcenko:
@xgui3783 @skoehnen 
I pushed a version with a smaller base image. Could you do some quick testing if everything works for you?

xgui3783:
size is definitely a little smaller: https://docker-registry.ebrains.eu/harbor/projects/28/repositories/siibra-python

base image | inflated | gzipped
--- | --- | ---
jupyter/scipy-notebook | 2.86GB | 895.42MB
jupyter/minimal-notebook | 2.67GB | 786.34MB

I will try to report on the functionalities when/if I get a chance

PR #151, opened 2022-01-11
First Version of spack config

marcenko:
I started to implement a first Version of the spack configuration.

It is not yet tested, since spack is not supported by windows.


xgui3783:
re: renaming `ebrains/spack` -> `.ebrains/spack`, emailed thanasis on 2022 03 02

marcenko:
Response to renaming:
It will be possible to use `.ebrains/spack`
But this method is still WIP.

PR #158, opened 2022-02-22
feature: addition to pydantic models

xgui3783:
add additional fields necessary for functioning of siibra-explorer


xgui3783:
@marcenko , can you take a look at this PR? It goes hand in hand with https://github.com/FZJ-INM1-BDA/siibra-api/pull/73 (I will fix the test on siibra-api)

marcenko:
@xgui3783 Sure. Was already looking into FZJ-INM1-BDA/siibra-api#73. Since they relate I will do both.

PR #166, opened 2022-03-11
reduce number of features for connectivity test

marcenko:
@xgui3783 
I saw that the tests are running over 1 hour. The connectivity tests are the one that cumsumes most of the time.
Does it make sense to run the tests over all features? I changed it to 10 for each Query. Now the pipeline is done in 12 minutes.

### DIFF
--- a/test/features/test_connectivity.py
+++ b/test/features/test_connectivity.py
@@ -28,7 +28,7 @@ queries_tuple = (
 queries_features_tuple = (
     feat
     for Query in queries_tuple
-    for feat in Query().features
+    for feat in Query().features[:10]
 )
 
 @pytest.mark.parametrize('conn_feat', queries_features_tuple)


xgui3783:
my concern is some of the features are malformed, and cannot be properly serialized (which is already true for a number of connectivity features. 

Maybe we can split the tests, split by a flag. 

- quick_flag is set, run on every push every branch, only test the serialization of the first 10 features/parcellations/regions etc
- quick_flag is unset, run on every push to main only, runs all tests
-

marcenko:
- Did some changes and added a way to define a fast run option for tests.
- The Quick run option is set for feat branches

xgui3783:
rather than `*feat*` running fast, I was thinking any github ref **except** `main` runs fast

This means all PR will have a fast CI check, but before we publish (on main, presumably), it will always run a full test?


also, why are we skipping python 3.7 oO?

marcenko:
Having the full run only on main makes sence. I will change it.

I removed 3.7 from the matrix because the tests are running again while creating the coverage report. So we were running the tests for 3.7 twice.

xgui3783:
> I removed 3.7 from the matrix because the tests are running again while creating the coverage report. So we were running the tests for 3.7 twice.

I see. 

Hmm, from quick glance perspective, I think it still makes sense to keep 3.7 in test phase, but I do agree that the tests are some what duplicated. 

How about we code coverage we always run in short mode? This way, codecov doesn't run forever, but python 3.7 is tested explicitly. 

The other reason is that codecov CI can fail when:

- codecov report upload was unsuccessful (al beit rare), and 
- in future, we might want to consider fail CI if codecov falls below a threshold

marcenko:
The changes are in. 
- 3.7 is back in test matrix
- coverage is running always fast
- only main runs with full tests

~ 12 minutes for the fast tests seems fine for the moment. Maybe we can optimize it in the future 😄

.github/workflows/siibra-testing.yml
```diff
@@ -33,7 +34,13 @@ jobs:
     - name: Install test dependencies
       run:  pip install pytest pytest-cov coverage
     - name: Run test with pytest
-      run: pytest -rx
+      run: |
+        echo "Using github.ref: $GITHUB_REF"
+        if [[ "$GITHUB_REF" != *"main"* ]]
```
xgui3783, line 39: I'm a bit rusty in bash, but this would match if github_ref has main anywhere?

perhaps we could use:

```
if [[ "$GITHUB_REF" != "refs/heads/main" ]]
```

? Otherwise looks great! Thanks!

xgui3783: @marcenko since it's such a small change, would you accept if I change it quickly, and merge this PR?

marcenko: @xgui3783 Sure. I'm absolutely fine with it

PR #168, opened 2022-03-19
 Store point and pointSet as annotations

fsdavid:


dickscheid:
directly connected to #169

xgui3783:
I suggest we close this and keep discussion at #262

PR #169, opened 2022-03-21
Create user class to store annotations

fsdavid:


dickscheid:
@AhmetNSimsek lets discuss in an upcoming developer meeting wether the drive connector should be re-integrated for v0.4+. It never went into a working version so far.

fsdavid:
Keep issue open - https://github.com/FZJ-INM1-BDA/siibra-python/issues/262

dickscheid:
If we introduce this, should be implemented at the level of the base "Location" class, using json serialization of the derived objects

xgui3783:
I suggest we close this PR and keep discussing on #262

PR #173, opened 2022-03-24
model fix: refining region internal identifier

xgui3783:


xgui3783:
superceded by #174 

repush to trigger CI

PR #176, opened 2022-03-28
Feat cell density to model

marcenko:


marcenko:
@xgui3783 I started to add a to model function for CellDensity so I can use it in siibra-api.

Could you have a look at it and tell wether that is the data you need for the explorer or what needs to be added?

siibra/features/cells.py
```diff
@@ -18,16 +18,28 @@
...
 import importlib
 import pandas as pd
 import nibabel as nib
+from pydantic import Field
 
-class CorticalCellDistribution(RegionalFeature):
+
+class CorticalCellDistributionModel(ConfigBaseModel):
+    id: str = Field(..., alias="@id")
+    type: str = Field('siibra/features/cells', const=True)
```
xgui3783, line 36: Ideally, we can can use `@type` as an alias for type?

siibra/features/cells.py
```diff
@@ -185,6 +198,21 @@ def plot(self, title=None):
         view.add_markers([tuple(self.location)])
         return fig
 
+    @property
+    def model_id(self):
+        return super().model_id
+
+    def to_model(self, detail=False, **kwargs) -> CorticalCellDistributionModel:
+        base_dict = dict(super().to_model(detail=detail, **kwargs).dict())
+        base_dict["type"] = "siibra/features/cells"
```
xgui3783, line 207: similarly, per above, if we could use `@type`  instead of `type` ?

siibra/features/cells.py
```diff
@@ -185,6 +198,21 @@ def plot(self, title=None):
         view.add_markers([tuple(self.location)])
         return fig
 
+    @property
+    def model_id(self):
+        return super().model_id
+
```
xgui3783, line 204: If I remember correctly, per PR https://github.com/FZJ-INM1-BDA/siibra-python/pull/170 , there may need to be a class method called `get_model_type`. 

In its absence, the super class's `get_model_type` will be used (in this case, it would be `Dataset` )

siibra/features/cells.py
xgui3783: (I know I added code like this else where, too), but since the introduction of `get_model_type` class method, writing model type, IMO, should be:

```python
base_dict['@type'] = self.get_model_type()
```

xgui3783 (changes requested):
Thanks for working on this!

In addition to the comments, do you think we could implement tests for cell density, similar to, e.g. receptor (https://github.com/FZJ-INM1-BDA/siibra-python/blob/main/test/features/test_receptors.py#L51-L56)

marcenko:
Added the changes and some tests

xgui3783:
Thanks!

PR #202, opened 2022-07-09
Fix groupregion decoding & centroid computation

dickscheid:
This PR introduces a few fixes to correct centroid computation and region decoding for connectivity matrices (in the HCP connectivity datasets by Domhof et al). In particular, it handles group regions in more in a more stable fashion when decoding from region specs, and when computing masks. 

Spatial computations in the Region object should be refactored in a future version, and left to the parcellationmap objects. I did not introduce such refactoring here to move on quickly, since the Showcase 1 work in HBP's WP1 requires these fixes soonish.

### DIFF
--- a/siibra/core/parcellation.py
+++ b/siibra/core/parcellation.py
@@ -330,7 +330,7 @@ class Parcellation(
     def publications(self):
         return self._publications + super().publications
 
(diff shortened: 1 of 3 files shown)


xgui3783 (commented):
Looks good so far. 

Would be better if we add some tests to ensure the functionality is stable. 

Can you provide some examples of the usage? I can add them to the tests.

PR #204, opened 2022-07-19
feat: device flow user auth

xgui3783:


xgui3783:
@dickscheid and @skoehnen I just tested the device flow again, and it seems to work. 

Axel set the polling time to 5 sec, so our client will poll the token endpoint once every 5 seconds to see if the user authenticated.

PR #210, opened 2022-08-04
feat: use separate volumes

xgui3783:


xgui3783:
This PR will be reworked based on how parcellation map is refactored in configuration

PR #220, opened 2022-09-29
fix lambda functions of continuous maps

dickscheid:
Some lambda functions for the cached maploaders of ContinuousParcellationMap were missing default argument definitions, which lead to an error when applying fetch_iter() on probability maps. This PR should fix that problem.


xgui3783:
Looks good to me.

If urgent, feel free to merge right away.

If not, ideal if we can add tests to prevent regression.

PR #226, opened 2022-11-14
WIP migrating pydantic models

xgui3783:


dickscheid:
Looks good. I didn't check all the features/ files, since the followup work on new configuration schemes will include a thorough review of the features classes, before we deliver the next siibra-python package to users.

PR #228, opened 2022-12-02
fix use configuration

xgui3783:
fix misc typing


dickscheid:
I merge this to facilitate development, but we have to revisit the order of module imports - the env variable solution should ultimately not be necessary in my eyes.

PR #229, opened 2022-12-05
Refactor v0.4 fix tests

xgui3783:
depends on https://github.com/FZJ-INM1-BDA/siibra-python/pull/228


siibra/features/voi.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, location, name):
 
     @classmethod
     def _from_json(cls, definition):
```
dickscheid, line 38: the feature _from_jsons will also go to the factory, ultimately classes in core/ and features/ should only access REGISTRY if needed, but not Factory, so that we have one place to implement and optimize predefined object construction.

dickscheid: I will merge this but work on this very soon in the v0.4 branch.

siibra/volumes/map.py
```diff
@@ -196,7 +198,7 @@ def space(self):
     def parcellation(self):
         for key in ["@id", "name"]:
             if key in self._parcellation_spec:
-                return REGISTRY.Parcellation[self._parcellation_spec[key]]
+                return REGISTRY[Parcellation][self._parcellation_spec[key]]
```
dickscheid, line 201: if only used for accessing the registry, we might use a string to save the extra import (here: "Parcellation" str instead Parcellation type)

siibra/volumes/parcellationmap.py
```diff
@@ -1045,50 +1045,8 @@ def __init__(self, parcellation, space):
...
-                    ] = r
-                    mapindex += 1
-
-        # either load or build the sparse index
-        if not self._load_index():
-            self._build_index()
-            self._store_index()
-        assert self.spatial_index.max() == len(self.probs) - 1
+        # TODO check if this functionality has been ported to elsewhere.
+        raise NotImplementedError(f"ImageProvider has been deprecated.")
```
dickscheid, line 1049: this is an orphan, not needed any more, I agree

siibra/volumes/map.py
xgui3783: I see your concern. 

I can address both of our concerns, I believe, in a future commit.

xgui3783:
ping @dickscheid , can you take a look at this PR? 

I would love merge this, and continue to add the tests. 

Several things to note:

- Please note that this drops support for 3.6, since it is EOL on Dec2021, and github action seems to no longer support it. 
- only atlas is tested at the moment. I will continue to add tests in next few days.

PR #231, opened 2022-12-06
bugfix: Parcellation.supported_spaces crash

### LINKED ISSUE #230: Getting spaces from Julich parcellation kills python kernel on refactor_v0.4a8
The following code
```
import siibra
siibra.parcellations["julich"].spaces
```
caused kernel to crash:
"Canceled future for execute_request message before replies were done
The Kernel crashed while executing code in the the current cell or a previous cell."

xgui3783:
fixes #230

### DIFF
--- a/siibra/core/parcellation.py
+++ b/siibra/core/parcellation.py
@@ -182,7 +182,10 @@ class Parcellation(Region):
         """
         return list(
             set(super().supported_spaces)
-            | {space for region in self for space in region.supported_spaces}
+            | { space
+                for region in self
+                if region is not self
+                for space in region.supported_spaces }
         )
 
     def supports_space(self, space: Space):


siibra/core/parcellation.py
```diff
@@ -182,7 +182,10 @@ def supported_spaces(self) -> Set[Space]:
         """
         return list(
             set(super().supported_spaces)
-            | {space for region in self for space in region.supported_spaces}
+            | { space
+                for region in self
+                if region is not self
+                for space in region.supported_spaces }
```
dickscheid, line 188: a more elegant way would be `for region in self.descendants`, which would exclude self, can you try?

dickscheid:
In fact, a working version was already implemented in Region, but Parcellation (as a special region) still had an obsolete implementation of supported_spaces, which I just removed to solve the problem.

PR #236, opened 2023-01-06
UnicodeEncodeError fix (issue for Windows)

AhmetNSimsek:
On Windows, the following code produces "UnicodeEncodeError : 'charmap' codec can't encode character '\u200b' in position 177 in ebrains.py : character maps to <undefined>"

```python
import siibra
from os import environ
environ['SIIBRA_ENABLE_DEVICE_FLOW'] = "1"
siibra.fetch_ebrains_token()

v2 = siibra.get_region("julich 2.9", "v2")
features = siibra.features.get(v2, siibra.features.external.EbrainsDataset)
for feature in features:
    print(f" - {feature.name}")
```

Adding the `encoding='utf-8'` in NamedTemporaryFile at line 115  in ebrains.py fixes the issue.

PS: '\u200b' refers to zero width space


xgui3783:
can you add a e2e test to prevent regression? (you can refer to https://github.com/FZJ-INM1-BDA/siibra-python/blob/refactor_v0.4/e2e/features/connectivity/test_connectivity.py on how tests are written and https://github.com/FZJ-INM1-BDA/siibra-python/blob/refactor_v0.4/.github/workflows/siibra-testing.yml#L47-L78 on how tests are run.

we can potentially add windows as a target os in our CI.

PR #237, opened 2023-01-09
Added a get_region_colormap method to be able to show selected region in color

AhmetNSimsek:
A method suggestion, based on `get_colormap()`, to show fetched regions and meshes in color when available. Could be useful when displaying the full map and a few individual regions.

Example code
```python
import siibra
import numpy as np
from nilearn import plotting
m = siibra.get_map(parcellation="v2.9", space="MNI colin 27")
spec = m.regions[np.random.randint(0, len(m.regions) - 1)]
print(spec)
region_cmap = m.get_region_colormap(spec)
mesh = m.fetch(region=spec, format="mesh")
plotting.view_surf((mesh["verts"], mesh["faces"]), cmap=region_cmap )
```

### DIFF
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -532,6 +532,28 @@ class Map(_concept.AtlasConcept, configuration_folder="maps"):
         ) / [255, 255, 255, 1]
         return ListedColormap(pallette)
 
+    def get_region_colormap(self, regionname):
+        """Generate a matplotlib colormap from known rgb values of label indices for a given region."""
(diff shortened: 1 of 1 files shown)


xgui3783:
since the implementation is almost identical to that of `get_colormap`, I wonder if we should add an optional `regions` kwarg to `get_colormap`. If set, will filter the regions based on provided value. If unset, will return listcolormap of all color maps?

AhmetNSimsek:
That is a good point; it would be better that way. I'll make a version and commit that to this branch too.

PR #246, opened 2023-01-19
Simplify device flow for notebooks

AhmetNSimsek:
```python
import siibra
siibra.fetch_ebrains_token(enable_flow=True) # or siibra.fetch_ebrains_token(True)
```

### DIFF
--- a/siibra/retrieval/requests.py
+++ b/siibra/retrieval/requests.py
@@ -261,14 +261,19 @@ class EbrainsRequest(HttpRequest):
             logger.warn("expected device_authorization_endpoint in .well-known/openid-configuration, but was not present")
 
     @classmethod
-    def fetch_token(cls, enable_flow: bool = False):
+    def fetch_token(cls):
         """Fetch an EBRAINS token using commandline-supplied username/password
         using the data proxy endpoint.
         """
-        if enable_flow:
-            from os import environ
-            environ["SIIBRA_ENABLE_DEVICE_FLOW"] = "1"
         cls.device_flow()
+
+    @classmethod
+    def is_jupyter_notebook(cls):
+        """
+        Is the code running in the jupyter environment.
+        Includes -notebook, -lab, vscode notebook, and google colab notebooks. 
+        """
+        return True if any(k in ['JPY_INTERRUPT_EVENT', "JPY_PARENT_PID"] for k in os.environ) else False
 
     @classmethod
     def set_token(cls, token):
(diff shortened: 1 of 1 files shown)


xgui3783:
The fetch token method should be able to tell a user environment vs machine environment.

I hit a bump and wasn't able to distinguish notebook environment, but this could be useful : https://github.com/fastai/fastprogress/issues/26#issuecomment-464951467

siibra/retrieval/requests.py
```diff
@@ -261,22 +261,29 @@ def init_oidc(cls):
...
         cls.device_flow()
 
+    @classmethod
+    def is_jupyter_notebook(cls):
+        """
+        Is the code running in the jupyter environment.
+        Includes -notebook, -lab, vscode notebook, and google colab notebooks. 
+        """
+        return True if any(k in ['JPY_INTERRUPT_EVENT', "JPY_PARENT_PID"] for k in os.environ) else False
+
```
xgui3783, line 277: probably doesn't need to be a whole function... see below

siibra/retrieval/requests.py
```diff
@@ -261,22 +261,29 @@ def init_oidc(cls):
...
     @classmethod
     def set_token(cls, token):
         logger.info(f"Setting EBRAINS Knowledge Graph authentication token: {token}")
         cls._KG_API_TOKEN = token
 
     @classmethod
     def device_flow(cls):
+        if cls.is_jupyter_notebook():
+            os.environ["SIIBRA_ENABLE_DEVICE_FLOW"] = "1"
```
xgui3783, line 286: rather than check is notebook here... (see below)

AhmetNSimsek:
> 

Thank you for the notes and the suggestion. I have committed a revision: now Siibra automatically checks if there is a Jupyter environment running including JupyterLab, VSCode Jupyter-Notebook, and Google Colab Notebooks.
The user only needs to use:
```python
siibra.fetch_ebrains_token()
```
as it would be in other environments.

siibra/retrieval/requests.py
```diff
@@ -261,22 +261,29 @@ def init_oidc(cls):
...
 
         if not sys.__stdout__.isatty() and not os.getenv("SIIBRA_ENABLE_DEVICE_FLOW"):
```
xgui3783, line 288: add check is jupyter environment here...

in principle, this is where fail fast is checked (can be overwritten with env var)

xgui3783 (commented):
some minor refactoring. otherwise, looks good

Issue #240, opened 2023-01-16
[0.4][feature request] get_region should support regex syntax (e.g. `/[expression]/flag`) 


xgui3783:
for refactor 0.4a

xgui3783:
fixed by #251

PR #251, opened 2023-01-23
regex syntax support for region.find(), i.e.,  '/pattern/flags'

AhmetNSimsek:
Closes issue #240:
It is sufficient to make `region.find()` support regex syntax to provide it for `parcellation.get_region()`.

Example:
```python
import siibra
import re
julich_brain = siibra.parcellations.get('julich 2.9')
# using string in '/pattern/flags' format (flags: not case-sensitive and only takes valid options).
print(julich_brain.find('/hOc[24].*right/iS'))
# using regex applied input
print(julich_brain.find(re.compile('hOc[24].*right', flags=re.S | re.I)))
```

### DIFF
--- a/siibra/core/region.py
+++ b/siibra/core/region.py
@@ -49,6 +49,9 @@ class Region(anytree.NodeMixin, concept.AtlasConcept):
     Representation of a region with name and more optional attributes
     """
 
+    _regex_re = re.compile(r'^\/(?P<expression>.+)\/(?P<flags>[a-zA-Z]*)$')
+    _accepted_flags = "aiLmsux"
+
     def __init__(
         self,
         name: str,
@@ -106,7 +109,7 @@ class Region(anytree.NodeMixin, concept.AtlasConcept):
         )
         self._supported_spaces = None  # computed on 1st call of self.supported_spaces
         self._CACHED_REGION_SEARCHES = {}
-
+        
     @property
     def id(self):
         if self.parent is None:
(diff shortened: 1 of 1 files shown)


xgui3783:
hmm, I wonder if using a regex pattern is easier, rather than manually search for flags etc:

```python
import re

regex_string = r'^\/(?P<expression>.+)\/(?P<flags>[aiLmsux]*)$'
regex = re.compile(regex_string)
match = regex.search('/hoc[0-9]/i')
assert match
print(match.group('expression'))
print(match.group('flags'))
```

siibra/core/region.py
```diff
@@ -218,11 +219,23 @@ def find(
...
+                import re
+                regex_specs = regionspec.split("/")
+                options = regex_specs[-1] if regex_specs[-1] != '' else None
+                flagmembers = re.RegexFlag.__members__
+                regionspec = re.compile(
+                    regex_specs[1],
+                    flags=0 if options is None else sum(
+                        [flagmembers[k.upper()].value for k in options if k.upper() in flagmembers]
+                        )
+                    )
```
xgui3783, line 238:
```suggestion
        if isinstance(regionspec, str):
            if regionspec in self.names:
                # key is given, this gives us an exact region
                match = anytree.search.find_by_attr(self, regionspec, name="key")
                MEM[key] = [] if match is None else [match]
                return MEM[key]
            
            # perhaps move to global, so it does not have to assign every request
            import re
            regex_string = r'^\/(?P<expression>.+)\/(?P<flags>[a-zA-Z]*)$'
            regex_re = re.compile(regex_string)
            accepted_flags = "aiLmsux"
            # end perhaps move to global
            
            regex_match = regex_re.match(region_spec)
            if regex_match:
                flags = regex_match.group('flags')
                expression = regex_match.group('expression')
                
                flag_string = ''
                for flag in flags or []: # catch if flags is nullish
                    if flag not in accepted_flags:
                        raise Exception(f"only accepted flag are in { accepted_flags }. {flag} is not within them")
                search_regex = (f"(?${flags})" if flags else "") + expression
                regionspec = re.compile(search_regex)
```

AhmetNSimsek: Good point. Pushed a commit (ead59a5e4906fcaa5399e1f1074275aeb318d51b) based on your suggestions.

PR #252, opened 2023-01-24
Refactor v0.4 suggestions for feature examples

AhmetNSimsek:
This branch includes some suggestions to enrich the examples of data features and add plot_matrix() method to display heatmaps of connectivity matrices.


examples/03_data_features/003_cell_distributions.py
```diff
@@ -40,20 +40,33 @@
...
+# We can, for example, plot the 2D distribution of the cell locations colored by layers:
 plt.scatter(c["x"], c["y"], c=c["layer"], s=0.2)
 plt.title(f"Cell distributions in {v1.name}")
 plt.grid(True)
 plt.axis("equal")
 plt.tight_layout()
 
+# %%
+# Having the data in data frame format allows further flexibility such as:
+layer1_cells = c[c["layer"]==1]
```
dickscheid, line 57: pandas provides even cleaner ways for this: `c[c.layer == 1]` or, which I prefer much,  `c.query('layer == 1')`

examples/03_data_features/003_cell_distributions.py
```diff
@@ -40,20 +40,33 @@
...
+plt.scatter(
+    layer1_cells["x"], layer1_cells["y"],
```
dickscheid, line 59: same here, layer1_cells.x etc.

examples/03_data_features/003_cell_distributions.py
```diff
@@ -40,20 +40,33 @@
...
+    s=layer1_cells["area(micron**2)"], c=layer1_cells["label"]
+    )
+mean_cell_area_layer_1 = layer1_cells["area(micron**2)"].mean()
```
dickscheid, line 62: just a style question, but I would avoid the extra line in this case, and put the mean() statement right in the f-string below - it is not much longer anyways.

examples/03_data_features/006_connectivity_matrices.py
```diff
@@ -59,7 +58,15 @@
 # with region objects as index.
 subject = conn.subjects[0]
 matrix = conn.get_matrix(subject)
-matrix
+matrix.head()
```
dickscheid, line 61: what is this line doing?

siibra/features/basetypes/regional_connectivity.py
```diff
@@ -131,6 +131,43 @@ def get_matrix(self, subject: str = None):
...
+        regions: list[str]
+            Display the matrix only for selected regions. By default, shows all the regions.
+            It can only be a subset of regions of the feature.
+        logscale: bool
+            Whether to show the data in log10 scale or not.
+        """
+        matrix = self.get_matrix(subject=subject)
+        
+        if logscale:
+            from numpy import log10
```
dickscheid, line 153: isn't numpy already imported as np? I'd guess so.

siibra/features/basetypes/regional_connectivity.py
```diff
@@ -131,6 +131,43 @@ def get_matrix(self, subject: str = None):
...
+            matrix = log10(matrix)
+        if regions is None:
+            # TODO: Fix: self.regions is not equal to matrix.columns.to_list() due to space mismatch
+            # After the fix, update this with regions = self.regions
+            regions = matrix.columns.to_list()
+
+         # default kwargs
+        subject_title = subject if subject is not None else ""
```
dickscheid, line 161: short form is `subject_title = subject or ""` I believe (since None is not True)

siibra/features/basetypes/regional_connectivity.py
```diff
@@ -131,6 +131,43 @@ def get_matrix(self, subject: str = None):
...
+        kwargs["title"] = kwargs.get(
+            "title",
+            f"{subject_title} - {self.modality} in {', '.join({_.name for _ in self.anchor.regions})}"
+        )
+        kwargs["figure"] = kwargs.get("figure", (15, 15))
```
dickscheid, line 166: "figure" as a plot argument appears strange to me, do matplotlib or nilearn plots have such an argument?  what you provide, the tuple (15, 15); also looks more like figsize to me - for "figure" I would expect a figure object

dickscheid (changes requested):

siibra/features/basetypes/regional_connectivity.py
AhmetNSimsek: I was surprised as well but that is how nilern.plot_matrix lists the parameter. It can take `matplotlib.figure.Figure` or  `figsize :obj: tuple`. (There are dedicated methods within to use the correct type for correct usage.)

AhmetNSimsek (commented):
Pushed a commit with the changes and noted the use of "figure" in conversation.

siibra/features/basetypes/regional_connectivity.py
```diff
@@ -131,6 +131,42 @@ def get_matrix(self, subject: str = None):
             self._matrices[subject] = self._load_matrix(subject)
         return self._matrices[subject]
 
+    def plot_matrix(self, subject: str = None, regions: list[str] = None, logscale: bool = False, **kwargs):
```
xgui3783, line 134: this will need to be `List[str]`, or else it will fail in python < 3.10 (I think)

AhmetNSimsek: You are right, I will push a fix now

examples/03_data_features/003_cell_distributions.py
```diff
@@ -40,20 +40,33 @@
...
+# %%
+# We can, for example, plot the 2D distribution of the cell locations colored by layers:
+plt.scatter(c.x, c.y, c=c.layer, s=0.2)
 plt.title(f"Cell distributions in {v1.name}")
 plt.grid(True)
 plt.axis("equal")
 plt.tight_layout()
 
+# %%
+# Having the data in data frame format allows further flexibility such as:
```
dickscheid, line 56: good idea

dickscheid (approved):

PR #256, opened 2023-01-26
Enable getting randomized colormaps for compressed maps

AhmetNSimsek:
Compressing maps can lead to overlapping region labels and hence while `parcellationmap.compress` relabels the regions, `parcellationmap.get_colormap` may result in the same colors being used for different regions. This commit suggests a solution: generate a randomized colormap for compressed maps, which is saved locally and can be loaded next time `get_colormap` is invoked on the same compressed map.
Example:
```python
import siibra
from nilearn import plotting

jubrain_mpm = siibra.get_map(parcellation="julich 2.9", space="mni152", maptype="labelled")

jubrain_mpm_whole = jubrain_mpm.compress()

plotting.view_img(jubrain_mpm_whole.fetch(), cmap=jubrain_mpm_whole.get_colormap(), symmetric_cmap=False)

# repeat the plot to confirm the load mechanism
plotting.view_img(jubrain_mpm_whole.fetch(), cmap=jubrain_mpm_whole.get_colormap(), symmetric_cmap=False)

# It also supports coloring only the selected regions:
plotting.view_img(jubrain_mpm_whole.fetch(),
    cmap=jubrain_mpm_whole.get_colormap(["Area TPJ (STG/SMG) right", "Frontal-to-Temporal-I (GapMap) left"]),
    symmetric_cmap=False)
```

For discussion: While `_create_randomized_colormap` and `_load_randomized_colormap` are not user-end methods, it can potentially be employed by users. We can also create a `load_colormap()` for users' custom colormaps (might be particularly useful for colorblind users).

### DIFF
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -548,23 +548,28 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
         from matplotlib.colors import ListedColormap
         import numpy as np
 
-        colors = {}
         if region_specs is not None:
             include_region_names = { self.parcellation.get_region(region_spec).name for region_spec in region_specs }
         else:
             include_region_names = None
 
+        if "_compressed" in self.id:
+            clrs = self._load_randomized_colormap()
+        
+        colors = {}
         for regionname, indices in self._indices.items():
             for index in indices:
                 if index.label is None:
                     continue
-                    
                 if (include_region_names is not None) and (regionname not in include_region_names):
                     continue
                 else:
-                    region = self.get_region(index=index)
-                    if region.rgb is not None:
-                        colors[index.label] = region.rgb
+                    if "_compressed" in self.id:
+                        colors[index.label] = list(clrs[index.label-1])
+                    else:
+                        region = self.get_region(index=index)
+                        if region.rgb is not None:
+                            colors[index.label] = region.rgb
 
         pallette = np.array(
             [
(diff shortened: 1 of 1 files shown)


xgui3783:
The problem with this approach, as I see it, is that colour maps often are not made in the vacuum.

For example, Jülich brain hoc1 is always purple ish, and if it appears any other colour, it might be uncanny.

AhmetNSimsek:
> The problem with this approach, as I see it, is that colour maps often are not made in the vacuum.
> 
> For example, Jülich brain hoc1 is always purple ish, and if it appears any other colour, it might be uncanny.

I see what you mean. From this thought process, I had thought about an alternative version where the code employs a different shade/tone if there is a label clash. I've gone with this approach also considering coloring for maps with no pre-assigned colormaps. But this can be a contingency instead.
What do you think of the shade/tone-idea? (I can make changes accordingly and commit)

xgui3783:
> > The problem with this approach, as I see it, is that colour maps often are not made in the vacuum.
> > For example, Jülich brain hoc1 is always purple ish, and if it appears any other colour, it might be uncanny.
> 
> I see what you mean. From this thought process, I had thought about an alternative version where the code employs a different shade/tone if there is a label clash. I've gone with this approach also considering coloring for maps with no pre-assigned colormaps. But this can be a contingency instead. What do you think of the shade/tone-idea? (I can make changes accordingly and commit)

I think the following would be prudent:

- for jba 2.9 in mni152, generate plot, screenshot, and attach to this PR for the following:
  - before this PR
  - this PR
  - your new proposed solution (different shade/tone)

The first screenshot would hopefully demonstrate the problem, and last two screenshots would present two different solutions. When compared visually, it would be a/ alot easier to tell there is a probem and b/ decide on which is the more favoured solution, and in which way we can make improvements

AhmetNSimsek:
>     * for jba 2.9 in mni152, generate plot, screenshot, and attach to this PR for the following:
>       
>       * before this PR
jubrain_compressed_existing_colormap
>       * this PR
jubrain_compressed_random_colormap
>       * your new proposed solution (different shade/tone)
jubrain_compressed_existing_darkened_colormap

I do like the different shade version for compressed maps and if you also agree, I'll make a new commit with this version.

I think we should still leave the randomized colormap getter and load colormap as options for users when the map does not have a built-in color scheme. This way, they could employ different colormaps when the need arises. What do you think?

dickscheid:
I am not sure if I fully understand the problem behind these ideas.
In general, I would not implement much colormap functionality in siibra. There are other libraries that provide functionality to create random or high contrast colormaps. The idea in siibra was only to allow to retrieve a colormap of a supported parcellation, *if* the parcellation specification has colors defined. I would not recommend to add functionality to create new colormaps, since this is not really part of siibra's scope and we would need to maintain, document and discuss it without a real need.

If possible, a compressed map should provide a colormap that preserves the colors of the original parcellation map. IS there a case where this cannot be achieved by just remapping the original colormap to the new label indices?

Issue #245, opened 2023-01-18
[0.4] get functional connectivity matrix error


xgui3783:
commit: 3d1ceb496b8303760c3f481c5aa69d84584b051e

example code:

```python
import siibra
features = siibra.features.get(siibra.parcellations['2.9'], siibra.features.connectivity.FunctionalConnectivity)
assert len(features) > 0
feat = features[0]
feat.get_matrix(feat.subjects[0]) # error here
```

expected: returns matrix

actual: Runtimeerror

```
RuntimeError: Non-quadratic connectivity matrix 294x1 from 294-Julich-Brain/2FunctionalConnectivity/000/EmpCorrFC_concatenated.csv in ZipfileConnector: /home/xgui3783/.cache/siibra.retrieval/c5f0e85a1fc2657c64df393663e0dccc9c7da65fdfbc8d2ced6a5a181712b46e
```

AhmetNSimsek:
The issue is caused by the format of the data. The array comes in `nrows` with 1 column where each cell is a `str`. These strings are `nrow` many floats joined together with a space. I wrote a fix for when a subject is selected, however, when the mean is requested it still fails as it does not go through `_load_matrix` method. Currently, considering other possible fixes.

xgui3783:
This was actually the solution I was thinking about last night.

PR #259, opened 2023-01-30
Solves #245 by converting data stored in str to values

AhmetNSimsek:
Some connectivity data is delivered in arrays with rows of strings that are compromised of values combined with spaces. This change allows for handling this frequent exception. (Solves #245 )
Example code to test:
```python
import siibra
features = siibra.features.get(siibra.parcellations['2.9'], "connectivity")
m = []
for i, feat in enumerate(features):
    try:
        m.append(feat.get_matrix(feat.subjects[0]))
    except:
        print(i)
print(len(m), len(features))

m = []
for i, feat in enumerate(features):
    try:
        m.append(feat.get_matrix())
    except:
        print(i)
print(len(m), len(features))
```

### DIFF
--- a/siibra/features/basetypes/regional_connectivity.py
+++ b/siibra/features/basetypes/regional_connectivity.py
@@ -121,6 +121,15 @@ class RegionalConnectivity(Feature):
                         desc=f"Averaging {len(self)} connectivity matrices"
                     )
                 ]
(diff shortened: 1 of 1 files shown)


xgui3783:
oh, so it's actually space separated, rather than comma separated?

If that's the case, there is a much simpler solution. 

I would advise do not merge this PR for the moment.

PR #263, opened 2023-02-08
chore: clarified+unified anchor.regions types

xgui3783:
currently, anchor.regions can have `List[Region]` or `Dict[Region, AssignmentQualification]` types.

this ambiguity can lend itself to errors in the future. 

This PR unifies the return type, and always returns `Dict[Region, AssignmentQualification]` type


xgui3783:
will add test to prevent regression

PR #265, opened 2023-02-08
Temp fix to autodoc generation timeout problem

AhmetNSimsek:
Newly added voi features increase the doc generation time significantly. Removing voi-feature example from 03-000 to as a temporary fix to be able to produce docs.

### DIFF
--- a/examples/03_data_features/000_matchings.py
+++ b/examples/03_data_features/000_matchings.py
@@ -119,20 +119,7 @@ for f in siibra.features.get(occ, "ebrains"):
 # ^^^^^^^^^^^^^^^^
 #
 # There are also data features which are linked to locations in a specific reference space.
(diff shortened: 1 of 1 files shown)


xgui3783:
It seems 44f5eb44280682a2141b804f73700f81f9ead10a whilst attempting to address the issue, did not completely address the issue. 

see failed build at https://readthedocs.org/projects/siibra-python/builds/19431779/

xgui3783:
I think we can tentatively solve the issue by  merging this PR. 

However, build time & memory issue will continue to haunt us in the future, when/if we add more to sphinx gallery. 

To solve this one and for all, I wonder if we can use a dedicated runner to:

- run the examples (currently run by github action)
- build sphinx docs (currently run by RTD)

xgui3783 (approved):

PR #266, opened 2023-02-08
chore: minor rework on if's logic + typing

xgui3783:
minor refactor: feature.get_instances: collapsing nested if's by shortcircuiting

chore: minor typing, type checking


siibra/features/basetypes/volume_of_interest.py
```diff
@@ -88,6 +88,8 @@ def name(self):
 
     @property
     def boundingbox(self):
+        # Sanity check. What else would boundingbox property be, if not boundingbox?
```
dickscheid, line 91: the anchor location is always a bounding box, since VolumeOfInterest has its own implementation of Anchor (volumeOfInterestAnchor) - but why not check it...

PR #271, opened 2023-02-13
Disable progress bars (`tqdm`) with logger level QUIET

AhmetNSimsek:


xgui3783 (commented):
I think it is probably easier if we have a `disable_tqdm` varaible in `siibra.common` or `siibra.logger`

in the future, if we want to change the condition (e.g. use certain variable, or update the logger level from 20 to 40, we don't have to change all instances again.

AhmetNSimsek:
The Sphinx conf.py currently sets `SIIBRA_LOG_LEVEL` to `ERROR`, effectively making it run on siibra.QUIET. So if tqdms are disabled with `logger.level`, we do not need any further changes.

AhmetNSimsek:
It is decided that a progress bar using `tqdm` will be added in `commons`. All `tqdm`s then will be replaced by siibra progress bar which is disabled within, if the log level is set to ERROR (siibra.QUIET).

PR #275, opened 2023-02-16
Get connectivity features' resting-state paradigm from configs

AhmetNSimsek:
While the paradigm data is stored in the preconfigs, it was being replaced with the string "Resting State". The paradigm provides important information when it is provided, so this PR allows siibra to take it from the jsons.


xgui3783 (approved):
LGTM

PR #277, opened 2023-02-21
Add a progressbar function with the ability to disable from logger level

AhmetNSimsek:
`commons._progresbar()` is just `tqdm` that can be centrally disabled by the logger. The alternative, was to add `disable=logger.level>20` to all `tqdm`s, see #271, but it could be easily forgotten for new bars.

Moving forward, all new progress bars should be made with this function.

This allows the removal of progress bars from tutorials in the doc.


xgui3783 (commented):
Why don't we just call the export tqdm, rather than _progressbar?

Those who knows tqdm would immediately know how the function should be used.

xgui3783:
Otherwise, lgtm

AhmetNSimsek:
> Otherwise, lgtm

The purpose is to centralize the ability to disable with logger level. This is to prevent possibly forgetting to add `disable=logger.level>20` for every new bar.

xgui3783:
> > Otherwise, lgtm
> 
> The purpose is to centralize the ability to disable with logger level. This is to prevent possibly forgetting to add `disable=logger.level>20` for every new bar.

This I get. What I mean is, insteadof

```python
# in siibra/commons.py
from tqdm import tqdm

def _progressbar(*args, **kwargs):
    return tqdm(
        *args,
        disable=kwargs.pop("disable", False) or (logger.level > 20),
        **kwargs
        )
```

Why don't we have 

```python
# in siibra/commons.py
from tqdm import tqdm as _tqdm

def tqdm(*args, **kwargs):
    return _tqdm(
        *args,
        disable=kwargs.pop("disable", False) or (logger.level > 20),
        **kwargs
        )
```

so in other modules, they can simply have

```python
from ..commons import tqdm

for something in tqdm(iterable):
    pass
```

AhmetNSimsek:
Ah I see what you mean. I do not have a preference; we could definitely do that. I am not sure how can we future-proof the import of `commons.tqdm` vs `tqdm.tqdm`. Perhaps we could have it similar to `siibra.logger` by `siibra.tqdm`?

PR #278, opened 2023-02-22
Allow specifying a map candidate of the same kind

AhmetNSimsek:
With the inclusion of Julich brain 3.0, the possibility of multiple maps of the same space and type has risen. This commit informs the options if not specified, and gets the specified map.

### DIFF
--- a/siibra/core/parcellation.py
+++ b/siibra/core/parcellation.py
@@ -156,8 +156,10 @@ class Parcellation(region.Region, configuration_folder="parcellations"):
             Use MapType.STATISTICAL to request probability maps.
             Defaults to MapType.LABELLED.
         spec: str, optional
-            When there are multiple maps of the same type available for the same
-            reference space, select a specific option.
+            In case of multiple matching maps for the given parcellation, space
+            and type, use this field to specify keywords matching the desired
+            parcellation map name. Otherwise, siibra will default to the first
+            in the list of matches (and inform with a log message)
         Returns
         -------
         parcellationmap.Map or SparseMap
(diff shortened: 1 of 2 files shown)


siibra/core/parcellation.py
```diff
@@ -155,6 +155,9 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED):
             Type of map requested (e.g., statistical or labelled).
             Use MapType.STATISTICAL to request probability maps.
             Defaults to MapType.LABELLED.
+        spec: str, optional
```
dickscheid, line 158: this should be more specific. Suggestion "In case of multiple matching maps for the given parcellation, space and type, use this field to specify keywords matching the desired parcellation map name. Otherwise, siibra will default to the first in the list of matches (and inform with a log message)"

siibra/core/parcellation.py
```diff
@@ -175,7 +178,13 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED):
             logger.error(f"No {maptype} map in {space} available for {str(self)}")
             return None
         if len(candidates) > 1:
-            logger.warning(f"Multiple {maptype} maps in {space} available for {str(self)}, choosing the first.")
+            spec_candidates = [c for c in candidates if spec in c.name]
```
dickscheid, line 181: I usually test each word in spec, and require all to match. See e.g. in the region name search. would translate to something like this: `[c for c in candidates if all(w.lower() in c.name.lower() for w in spec.split())]`

siibra/core/parcellation.py
```diff
@@ -175,7 +178,13 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED):
...
+            if len(spec_candidates) > 1 or len(spec_candidates) == 0:
+                logger.warning(
+                    f"Multiple {maptype} maps in {space} available for {str(self)}, choosing the first."
+                    f"Set a unique `spec` from {[c.name for c in candidates]} to specify one."
```
dickscheid, line 185: better:"you might want to specify keywords using the spec parameter to match the desired one of these maps"

dickscheid (changes requested):
I suggest a few improvements

PR #280, opened 2023-02-23
bugfix: Image should not  have configuration folder + misc fixes

xgui3783:
bugfix: image should not have configuration_folder
test: add get all features test
chore: add typing
chore: generalize category as class var of Feature
chore: expose cache on siibra

### DIFF
--- a/e2e/features/test_get.py
+++ b/e2e/features/test_get.py
@@ -1,0 +1,8 @@
+import pytest
+import siibra
+
+# We get all registered subclasses of Feature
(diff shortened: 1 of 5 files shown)


AhmetNSimsek (approved):

AhmetNSimsek:
Should we perhaps also expose `from_json`, @xgui3783?

xgui3783:
I don't think user should access from_json. They should rely on use_configuration and extend_configuration instead.

cache is needed because they then can clear the cache

PR #282, opened 2023-02-23
feat: add ebrains ds to volumes

xgui3783:
Some critical metadata (e.g. description of pmap of juelich brain regions) are currently been associated with volume (e.g. https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/blob/master/maps/colin27-jba29-continuous.json#L11-14) 

Currently, ebrains property on `Volume` is not utilized (only classes subclassing `Atlasconcept` correctly interprets dataset. since `Volume` does not, it does not properly interprets dataset)

This PR fixes this issue, and volume will now properly have dataset property.


dickscheid (approved):
looks fine to me!

PR #285, opened 2023-02-24
Merge meshes from all labels

dickscheid:
If a neuroglancer mesh provider is linked to a parcellation map, and no label is given for fetching mesh fragments, collect them across all labels known in the parcellation instead of raising an error.

### DIFF
--- a/siibra/volumes/neuroglancer.py
+++ b/siibra/volumes/neuroglancer.py
@@ -535,7 +535,7 @@ class NeuroglancerMesh(volume.VolumeProvider, srctype="neuroglancer/precompmesh"
     def fetch(self, label: int, fragment: str):
         """
         Fetches a particular mesh. Each mesh is a dictionary with keys:
-        
+
         Parameters
         ----------
         label: int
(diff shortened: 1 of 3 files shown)


xgui3783:
I think I can understand where the problem lies. 

How would the user encounter this scenario? I can't help but to think that might be more elegant ways. 

For example, inside parcellationmap fetch method, when it calls volume's fetch method, it could pass it the exhaustive list of MapIndex.

siibra/volumes/neuroglancer.py
```diff
@@ -535,7 +535,7 @@ def _fetch_fragment(self, url: str, transform_nm: np.ndarray):
     def fetch(self, label: int, fragment: str):
```
xgui3783, line 535:
```suggestion
    def fetch(self, label: int, fragment: str, *, all_labels=None, **kwargs):
```

siibra/volumes/neuroglancer.py
```diff
@@ -556,37 +556,62 @@ def fetch(self, label: int, fragment: str):
...
-        fragment_infos = self._get_fragment_info(label)
+        if label is None:
+            if hasattr(self, "parcellation_map"):
+                # providers of a parcellation map volume are monkey-patched
+                # with a reference to the map. In that case, we can infer
+                # the set of available labels, but it cannot be inferred
+                # from the neuroglancer URL only.
+                labels = self.parcellation_map.labels
+            else:
+                raise RuntimeError(f"Fetching from {self.__class__.__name__} requires to specify a mesh label.")
```
xgui3783, line 567:
```suggestion
        if label is None:
            if all_labels is None:
                raise RuntimeError(f"Fetching from {self.__class__.__name__} requires to specify a mesh label.")
            # providers of a parcellation map volume are monkey-patched
            # with a reference to the map. In that case, we can infer
            # the set of available labels, but it cannot be inferred
            # from the neuroglancer URL only.
            labels = all_labels 

```

siibra/volumes/parcellationmap.py
```diff
@@ -157,6 +157,9 @@ def __init__(
         self._parcellation_spec = parcellation_spec
         self._affine_cached = None
         for v in self.volumes:
+            # allow the providers to query their parcellation map if needed
+            for p in v._providers.values():
+                p.parcellation_map = self
```
xgui3783, line 162:
```suggestion

```

xgui3783 (changes requested):
I couldn't suggest the following edit:

```python
# in Map.fetch
            result = self.volumes[mapindex.volume or 0].fetch(
                fragment=mapindex.fragment, label=mapindex.label, all_labels=self.labels, **kwargs
            )
```

Rather than adding parcellation_map as a property to volumeprovider, is it a better idea to pass Map.labels reference (the suggested edit) as a kwarg during runtime (when fetch is called), since all kwargs are passed down.

We could then avoid the strong coupling between volume provider and map. Volume provider is a lower level concept, which should not reference a higher level concept such as map.

xgui3783:
the commits is already merge into main by way of https://github.com/FZJ-INM1-BDA/siibra-python/commit/d66c62ca467037c128551ec923eb8c8ea1551386 ?

so perhaps we can close this PR?

dickscheid:
I am not really happy with any of those. It is somewhat a flow in the creation of the neuroglancer meshes. I think there should be a json on the neuroglancer side which allows to read out the labels available there (without having access to get a director listing from the server). Isn't there such a thing?

AhmetNSimsek:
@dickscheid and @xgui3783, is it safe to assume, this PR is not relevant for version 1 anymore and it will be revisited along with the refactor of how meshes are handled in siibra-python?

Issue #269, opened 2023-02-10
[bug] regional connectivity have same id's


xgui3783:
commit: f43192d38e72e15ae5465bad3baeb92ac8400c7e

to reproduce:

```python
import siibra
features = siibra.features.get(siibra.parcellations['2.9'], "RegionalConnectivity")
assert len(features) == len({ f.id for f in features }), f"{len(features)} features, but only {len({ f.id for f in features })} unique id's"
```

AhmetNSimsek:
The data is different and the issue lies in how `feature.id` and `feature.name` are defined. Can be solved by a more descriptive `feature.name`.

xgui3783:
To raise some ideal property of the `id` property (pun intended)

- must be unique
- must be immutable (i.e. cannot use memory address, cannot use array/list index)
- must be quick to compute (i.e. cannot be computed by hashing underlying data, which are often network bound)

xgui3783:
use file hash to compute feature id

the file hash has already been added to snapshot, but integrating into feature id still needs to be implemented

PR #288, opened 2023-02-28
Fix: dedup feature_id and feature instances

xgui3783:
fix: Feature.match should return unique feature instances
fix: regional connectivity should return unique feature id

fixes https://github.com/FZJ-INM1-BDA/siibra-python/issues/269

### DIFF
--- a/e2e/features/connectivity/test_connectivity.py
+++ b/e2e/features/connectivity/test_connectivity.py
@@ -7,13 +7,27 @@ from e2e.util import check_duplicate
 
 jba_29 = siibra.parcellations["2.9"]
 
-
-# FIXME
-@pytest.mark.xfail(reason="Some regional connectivity have same id. Remove mark as xfail when fixed.")
 def test_id_unique():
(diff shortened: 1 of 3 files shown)


AhmetNSimsek:
I think we should extend this to all features that have a paradigm by allowing the `id` property of `Feature` class to take `paradigm` if available.

xgui3783:
> I think we should extend this to all features that have a paradigm by allowing the `id` property of `Feature` class to take `paradigm` if available.

The problem is that `paradigm` is not a property of Feature. 

I would agree if we add paradigm as a property to Feature base class, which is not applicable to other Feature.

xgui3783:
> I think we should extend this to all features that have a paradigm by allowing the `id` property of `Feature` class to take `paradigm` if available.

I think we can consider adding paradigm as a property to all features at a later date. preferably, we should discuss it in a discussion thread or an issue. 

I would merge this for now, since it's a blocking issue for siibra-api.

PR #293, opened 2023-03-07
`Map.colorize` should take kwargs to allow fetching with further specification

AhmetNSimsek:
Particularly if one wants to fetch higher-resolution volumes.


xgui3783 (approved):
Lgtm

PR #303, opened 2023-03-16
chore: update schema

xgui3783:


AhmetNSimsek:
LGTM

AhmetNSimsek:
features/timseries as config folder is not the best option. Timeseries features will be divided into two: activity and image(fMRI). While activity is tabular image is not. These should be disinguished. Within siibra hierarchy this is easy as you had mentioned: `siibra.tabular.timeseries` but config should reflect the difference.

xgui3783:
So `timeseries/activity` for now?

Issue #187, opened 2022-05-15
[bug] parcellation find_region `area 4` returns empty suggestion


xgui3783:
as of c93b87c99e79a5a4a213f3f55a57020a82cb0c32 

to reproduce:

```python
import siibra
p = siibra.parcellations['2.9']
p.find_regions('area 4')
```

expected return:

```python
[area4p, area4a, area44]
```

actual return

```python
[]
```

AhmetNSimsek:
This bug has evolved:
```
[Anterior area, layer 4,
 Visceral area, layer 4,
 Postrhinal area, layer 4,
 Anteromedial visual area, layer 4,
 Medial visual area, layer 4,
 Rostrolateral area, layer 4,
 Anterolateral visual area, layer 4,
 Primary visual area, layer 4,
 Lateral visual area, layer 4,
 Dorsal auditory area, layer 4,
 Ventral auditory area, layer 4,
 Primary auditory area, layer 4,
 Posterior auditory area, layer 4,
 Laterointermediate area, layer 4,
 posteromedial visual area, layer 4,
 Primary somatosensory area, layer 4,
 Posterolateral visual area, layer 4,
 Entorhinal area, lateral part, layer 4,
 Supplemental somatosensory area, layer 4,
 Retrosplenial area, dorsal part, layer 4,
 Entorhinal area, lateral part, layer 4/5,
 Mediomedial anterior visual area, layer 4,
 Primary somatosensory area, nose, layer 4,
 Primary somatosensory area, trunk, layer 4,
 Mediomedial posterior visual area, layer 4,
 Rostrolateral lateral visual area, layer 4,
 Primary somatosensory area, mouth, layer 4,
 Laterolateral anterior visual area, layer 4,
 Primary somatosensory area, lower limb, layer 4,
 Primary somatosensory area, unassigned, layer 4,
 Primary somatosensory area, upper limb, layer 4,
 Primary somatosensory area, barrel field, layer 4,
 Entorhinal area, medial part, dorsal zone, layer 4,
 Entorhinal area, medial part, ventral zone, layer 4,
 Anterior area, layer 4,
 Visceral area, layer 4,
 Postrhinal area, layer 4,
 Anteromedial visual area, layer 4,
 Medial visual area, layer 4,
 Rostrolateral area, layer 4,
 Anterolateral visual area, layer 4,
 Lateral visual area, layer 4,
 Primary visual area, layer 4,
[...]

PR #320, opened 2023-04-03
Clarify the uses of `find` and `find_regions`

AhmetNSimsek:
Since `find_regions` is a class method of `parcellation` that goes through all the parcellations in the parcellation registry, when called from an instance, it also returns results from different parcellations (hence also species). Resolves the issue #187 by hiding the `find_regions` from the parcellation instance but making it available at `siibra.find_regions` and promoting the use of `find` method.

### DIFF
--- a/siibra/__init__.py
+++ b/siibra/__init__.py
@@ -47,7 +47,7 @@ logger.info(
 # forward access to some functions
 set_ebrains_token = _EbrainsRequest.set_token
 fetch_ebrains_token = _EbrainsRequest.fetch_token
-find_regions = _parcellation.Parcellation._find_regions
+find_regions = _parcellation.Parcellation.find_regions
 from_json = factory.Factory.from_json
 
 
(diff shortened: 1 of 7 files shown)


xgui3783 (commented):
is there a need to change the class method from `find_regions` to `_find_regions` ?

There doesn't seem to be any material changes?

AhmetNSimsek:
`find` and `find_regions` are not distinguishable by name in terms of functionality. But both are available from a parcellation instance. There is no reason for the user to choose one over the other and it will cause confusion. But if `find_regions` is not exposed from a parcellation instance, it is clear.
Do you have an alternative suggestion? We could potentially move `find_regions` out of parcellation class 🤔

AhmetNSimsek:
The examples already show the correct use and describe it. I think there is nothing really new about `region.find`. We had added the exact match part but now it is just taken out.
It now works as intended.

Unit tests, aside from e2e tests, wouldn't hurt.

xgui3783:
Aww, I didn't notice you actually added the usecases for regex. Good call!

PR #331, opened 2023-04-05
Clean region.find() and parcellation.get_region() up

### LINKED ISSUE #330: Cannot fetch colormap of the rat atlas
(as of 935503c)
```python
ratspace = siibra.spaces["waxholm"]
parc = siibra.parcellations["waxholm 4"]
mp = siibra.get_map(parc,ratspace)
mp.get_colormap()
```
This fails because the search for the structure "lateral olfactory tract" results in `["lateral olfactory tract", "Nucleus of the lateral olfactory tract"]` and `get_region` fails.

AhmetNSimsek:
Patches #330

### DIFF
--- a/e2e/core/test_region.py
+++ b/e2e/core/test_region.py
@@ -11,4 +11,16 @@ def test_region_spatial_props(region: siibra.core.parcellation.region.Region):
     props = region.spatial_props("mni152")
     for idx, cmp in enumerate(props.components, start=1):
         assert cmp.volume >= props.components[idx - 1].volume
-        
+
+
+regions_of_concern = [
+    ("julich 3.0", "frontal lobe"),
+    ("julich 1.18", "Ch 123 (Basal Forebrain) left"),
+    ("waxholm 4", "lateral olfactory tract")
+]
+
+# Test duplicate named regions and regions with only child
+@pytest.mark.parametrize("par_region_tuple", regions_of_concern)
+def test_get_region(par_region_tuple):
+    region = siibra.get_region(par_region_tuple[0], par_region_tuple[1])
+    assert region.name == par_region_tuple[1]
--- a/e2e/core/test_region.py
+++ b/e2e/core/test_region.py
@@ -1,5 +1,6 @@
 import pytest
 import siibra
+import re
 
 regions = [
     siibra.get_region("julich 3.0", "Area 4p (PreCG) right"),
@@ -20,8 +21,36 @@ regions_of_concern = [
 ]
 
 # Test duplicate named regions and regions with only child
-@pytest.mark.parametrize("parc_spec,region_name", regions_of_concern)
-def test_get_region(parc_spec,region_name):
+@pytest.mark.parametrize("parc_spec, region_name", regions_of_concern)
+def test_get_region(parc_spec, region_name):
     region = siibra.get_region(parc_spec, region_name)
     assert region
     assert region.name == region_name
+
+
+regionspecs = [
+    ("julich 3.0", "julich 3.0", 1,
+        [siibra.parcellations["julich 3.0"].name]
+    ),
+    ("waxholm 4", "lateral olfactory tract", 2,
+        ["lateral olfactory tract", "Nucleus of the lateral olfactory tract"]
+    ),
+    ("julich 2.9", "/area 4/i", 12,
+        ["Area 44 (IFG)", "Area 4p (PreCG) right", "Area 4a (PreCG) left"]
+    ),
+    ("Superficial fiber Bundles HCP", re.compile("rh_SF-SF_*"), 24,
+        ["rh_SF-SF_9", "rh_SF-SF_19", "rh_SF-SF_23"]
+    ),
+    ("julich 3.0", siibra.get_region("julich 3.0", "hoc1 left"), 1,
+        ["Area hOc1 (V1, 17, CalcS) left"]
+    )
+]
+@pytest.mark.parametrize(
+    "parc_spec, region_spec, result_len, check_regions", regionspecs
+)
+def test_get_region(parc_spec, region_spec, result_len, check_regions):
+    parc = siibra.parcellations.get(parc_spec)
+    results = parc.find(region_spec)
+    assert isinstance(results, list)
+    assert len(results) == result_len
+    assert all(r in results for r in check_regions)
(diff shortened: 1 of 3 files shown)


AhmetNSimsek:
Note: Remove the elimination of possible duplicates using `set` in `region.find` as it can remove regions with the same name but on a different hierarchy. However, it does check for exact match when filtering for the parent.
`parcellation.get_region()` is now responsible for sieving the results and allowing the exact match to be returned instead of throwing an error.

Example:
```python
import siibra
p = siibra.parcellations["waxholm 4"]
p.get_region("lateral olfactory tract")
```
was raising `RuntimeError: Spec lateral olfactory tract resulted in multiple matches: Nucleus of the lateral olfactory tract, Olfactory white matter.` but now returns `lateral olfactory tract`.

siibra/core/region.py
```diff
@@ -281,6 +274,7 @@ def find(
                         for r in filtered
                         if (r.parent is not None)
                         and all((c in filtered) for c in r.parent.children)
+                        and r.name != regionspec
```
xgui3783, line 277: why is it necessary to check this inequality?

siibra/core/region.py
```diff
@@ -306,7 +300,7 @@ def find(
         # reverse is set to True, since SequenceMatcher().ratio(), higher == better
         MEM[key] = (
             sorted(
-                set(found_regions),
+                found_regions,
```
xgui3783, line 303: I understand your sentiment, but what do we gain by removing set?

siibra/core/region.py
```diff
@@ -252,18 +252,11 @@ def find(
...
 
-            if regionspec in self.names:
-                # key is given, this gives us an exact region
-                match = anytree.search.find_by_attr(self, regionspec, name="key")
-                MEM[key] = [] if match is None else [match]
-                return list(MEM[key])
-
         candidates = list(
-            set(anytree.search.findall(self, lambda node: node.matches(regionspec)))
+            anytree.search.findall(self, lambda node: node.matches(regionspec))
```
xgui3783, line 256: I understand your sentiment, but what do we gain by removing set?

siibra/core/region.py
AhmetNSimsek: The thred that started all this was the region `lateral olfactory tract` in the rat parcellation. When fetching a colormap, `find` results as its parent because its parent has only one child. So the exact name check cannot be performed. This is a precaution against going upwards when there is an exact match.

siibra/core/region.py
AhmetNSimsek: If there are regions with the same name but at a different hierarchy like `frontal lobe` in JBA 3.0, they will be reduced to one. While it is best that the data doesn't have duplicate names, as it stands, `find` should find all matching the regionspec.

siibra/core/region.py
AhmetNSimsek: The same as above. I will conduct more tests to see if this regresses anything

siibra/core/parcellation.py
```diff
@@ -330,10 +326,22 @@ def get_region(
...
+            ]
+            if len(exact_matches) == 1:
+                return exact_matches[0]
+            else:
+                if len({r.parent for r in exact_matches}) == 1:
+                    logger.info(
+                        f"Found {len(exact_matches)} exact matches for "
+                        f"{regionspec} with the same parent. Returning the first."
+                        )
+                    return exact_matches[0]
```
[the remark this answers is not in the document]
AhmetNSimsek: The idea was to not reduce regions with parents with the same name.

AhmetNSimsek:
@xgui3783, I've converted back to a draft. It seems some more thought needs to be given to have it more concise and clear.
Also, I am not sure additional tests are necessary since when the initial correction was made, there were a few failing tests. However, I am considering adding a test for getting a colormap since a map (not all) should be able to find a unique region in the parcellation for all its regions.

siibra/core/parcellation.py
```diff
@@ -330,10 +326,22 @@ def get_region(
         elif len(candidates) == 1:
             return candidates[0]
         else:
+            exact_matches = [
+                r for r in candidates if r == regionspec or r.name == regionspec
+            ]
+            if len(exact_matches) == 1:
+                return exact_matches[0]
```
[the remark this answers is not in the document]
AhmetNSimsek: Good point

siibra/core/region.py
xgui3783: check with @dickscheid . I believe this is the intended behaviour, in different circumstances

AhmetNSimsek:
#335 summarizes all the issues.

siibra/core/region.py
AhmetNSimsek: https://github.com/FZJ-INM1-BDA/siibra-python/issues/335

AhmetNSimsek: Reintroduced set to avoid `siibra.parcellations["julich 1.18"].get_region("Ch 123 (Basal Forebrain) left")` (and similar scenarios) from failing. But now once again `siibra.parcellations["julich 3.0"].find('frontal lobe')` returns only the topmost 'frontal lobe'.

xgui3783 (approved):
LGTM

If memories serves, it was timo's idea where `julichbrain_29.get_region("hoc1")` should fetch the parent region. I was a never a big fan. 

I would add tests (ideally unit test, but we will do well to also add e2e tests) to ensure no regressions are made.

e2e/core/test_region.py
```diff
@@ -11,4 +11,16 @@ def test_region_spatial_props(region: siibra.core.parcellation.region.Region):
...
+    ("julich 3.0", "frontal lobe"),
+    ("julich 1.18", "Ch 123 (Basal Forebrain) left"),
+    ("waxholm 4", "lateral olfactory tract")
+]
+
+# Test duplicate named regions and regions with only child
+@pytest.mark.parametrize("par_region_tuple", regions_of_concern)
+def test_get_region(par_region_tuple):
+    region = siibra.get_region(par_region_tuple[0], par_region_tuple[1])
+    assert region.name == par_region_tuple[1]
```
xgui3783, line 26:
```suggestion
@pytest.mark.parametrize("parc_spec,region_name", regions_of_concern)
def test_get_region(parc_spec,region_name):
    region = siibra.get_region(parc_spec, region_name)
    assert region
    assert region.name == region_name
```

AhmetNSimsek:
> since significant changes were made to Region.find, perhaps it's prudent to add some tests there too.
> 
> What are the expected behaviors of Region.find etc

Added a few important use cases and checks.

PR #337, opened 2023-04-17
Add docstring to livequery

xgui3783:
chore: typo, simplify list


dickscheid:
much better readable now. Ok for me to merge. We will revisit this after profiling live queries, according to yesterday's discussion.

Issue #302, opened 2023-03-16
[bug] merged map do not have the same shape, or dtype


xgui3783:
as of 0.4a35

when merge labelled map, the returned nifti has the wrong shape and dtype

sample code:

```python
import siibra
labelled_map = siibra.get_map("2.9", "152", "labelled")
vol_l = labelled_map.fetch(fragment="left hemisphere")
vol_r = labelled_map.fetch(fragment="right hemisphere")
vol_b = labelled_map.fetch()

assert vol_l.dataobj.dtype == vol_r.dataobj.dtype == vol_b.dataobj.dtype # datatype differs (uint16, uint16, float64)
assert vol_l.dataobj.shape == vol_r.dataobj.shape == vol_b.dataobj.shape # shape differs ( (193, 229, 193), (193, 229, 193),  (289, 361, 271) )
```

AhmetNSimsek:
Note: `labelled_map.compress()` fetches the merged map as well and returns the correct shape. Looking into it.

AhmetNSimsek:
as of v0.4a41
`NiftiProvider._merge_fragments` creates a new nifti image before merging the fragments. During this, the dimensions are calculated via `voxdims = np.dot(np.linalg.inv(original_affine), np.r_[self.bbox.shape, 1])[:3]`.

Note: The merge is successful even when the original shape is taken.

Next: will compare the compressed map image and merged to see the exact difference

PR #341, opened 2023-04-19
Preserve map shape and dtype when a map is merged

AhmetNSimsek:
Solves #302

After review, the same should be applied to neuroglancer.

### DIFF
--- a/siibra/volumes/nifti.py
+++ b/siibra/volumes/nifti.py
@@ -94,7 +94,7 @@ class NiftiProvider(volume.VolumeProvider, srctype="nii"):
                 s0 = np.identity(4)
                 s0[:3, -1] = list(bbox.minpoint.transform(np.linalg.inv(img.affine)))
                 result_affine = np.dot(img.affine, s0)  # adjust global bounding box offset to get global affine
-                voxdims = np.r_[bbox.transform(result_affine).shape].astype(img.dataobj.dtype)
+                voxdims = np.asanyarray(bbox.transform(result_affine).shape, dtype="int")
                 result_arr = np.zeros(voxdims, dtype=img.dataobj.dtype)
                 result = nib.Nifti1Image(dataobj=result_arr, affine=result_affine, dtype=img.dataobj.dtype)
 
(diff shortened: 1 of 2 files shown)


siibra/volumes/nifti.py
```diff
@@ -95,9 +94,9 @@ def _merge_fragments(self) -> nib.Nifti1Image:
                 s0 = np.identity(4)
                 s0[:3, -1] = list(bbox.minpoint.transform(np.linalg.inv(img.affine)))
                 result_affine = np.dot(img.affine, s0)  # adjust global bounding box offset to get global affine
-                voxdims = np.dot(np.linalg.inv(result_affine), np.r_[bbox.shape, 1])[:3]
-                result_arr = np.zeros((voxdims + .5).astype('int'))
-                result = nib.Nifti1Image(dataobj=result_arr, affine=result_affine)
+                voxdims = np.r_[bbox.transform(result_affine).shape].astype(img.dataobj.dtype)
```
dickscheid, line 97: I don't understand this line:
- why are you using the np.r_ here?
- why would the voxel dimension be cast to float for a float image? A float image would still have integer shape (the intensities are float, but the pixel grid is always discrete in our applications here).

siibra/volumes/nifti.py
```diff
@@ -95,9 +94,9 @@ def _merge_fragments(self) -> nib.Nifti1Image:
                 s0 = np.identity(4)
                 s0[:3, -1] = list(bbox.minpoint.transform(np.linalg.inv(img.affine)))
                 result_affine = np.dot(img.affine, s0)  # adjust global bounding box offset to get global affine
-                voxdims = np.dot(np.linalg.inv(result_affine), np.r_[bbox.shape, 1])[:3]
-                result_arr = np.zeros((voxdims + .5).astype('int'))
-                result = nib.Nifti1Image(dataobj=result_arr, affine=result_affine)
+                voxdims = np.asanyarray(bbox.transform(result_affine).shape, dtype="int")
+                result_arr = np.zeros(voxdims, dtype=img.dataobj.dtype)
+                result = nib.Nifti1Image(dataobj=result_arr, affine=result_affine, dtype=img.dataobj.dtype)
```
dickscheid, line 99: I think the second type spec in line 99 can be left out, it should default to the array's dtype which is here obviously identical.

Issue #162, opened 2022-03-02
connectivity data feature bug: some col_names != row_names


xgui3783:
as of 0.3a16

to reproduce:

```
from siibra.features.connectivity import (
    PrereleasedStreamlineCountQuery,
)
features = PrereleasedStreamlineCountQuery().features
matrices = [feat.matrix for feat in features]
```

expected, matrices loads without issues

actual: several cannot be loaded due to

```
assertion error: expected col_names == row_names
```

Here is a printout of all datasets that trigger this issue:

```
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat3]
  reason: Averaged_SC_JuBrain_184Regions_HCP_10M_count_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat4]
  reason: Averaged_SC_JuBrain_184Regions_eNKI_10M_count_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat7]
  reason: Averaged_SC_JuBrain_184Regions_HCP_10M_length_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat8]
  reason: Averaged_SC_JuBrain_184Regions_eNKI_10M_length_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat11]
  reason: Averaged_FC_JuBrain_184Regions_HCP_REST_FIX_AVER_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat12]
  reason: Averaged_FC_JuBrain_184Regions_HCP_REST_FIX_CONCAT_MEAN assertion error: expected col_names == row_names
XFAIL test/features/test_connectivity.py::test_conn_to_model[conn_feat13]
  reason: Averaged_FC_JuBrain_184Regions_eNKI_REST_FIX_MEAN assertion error: expected col_names == row_names
```

PR #347, opened 2023-04-20
connectivity assert col_names == row_names (#162)

AhmetNSimsek:


xgui3783:
rather than add strict check in siibra, I think it's far more useful to add it as e2e test, so that when/if new datasets are added, violating this rule, it will be caught by CI, and not at runtime

AhmetNSimsek:
If someones uses their own data, then this check is still necessary imo.

xgui3783:
> If someone uses their own data, then this check is still necessary IMO. Can still add the tests as an extra but as long as get_matrix is tested, this will be included.

User will need to call `get_matrix` on every subject to see the error (unless I am mistaken?). When using custom data, they could run ci/cd to validate their data.

Not to mention, adding assertion at runtime taxes every call to `get_matrix`, whereas in reality, it really just need to be done once.

e2e/features/connectivity/test_connectivity.py
```diff
@@ -22,9 +22,13 @@ def test_connectivity_get_matrix(f: RegionalConnectivity):
     assert isinstance(f, RegionalConnectivity)
     assert len(f.subjects) > 0
     assert all(isinstance(subject, str) for subject in f.subjects)
+    matrix_df = f.get_matrix(subject=FileNotFoundError)
+    assert all(matrix_df.index[i] == r for i, r in enumerate(matrix_df.columns))
```
xgui3783, line 26: subject=filenotfound?

AhmetNSimsek: This is for testing the averaged matrix which requires `subject=None`.
We've recently made sure that this is the way instead using `subject='mean'`

AhmetNSimsek:
Note: The failing tests are due to bigbrain intensity and is not relevant.

e2e/features/connectivity/test_connectivity.py
xgui3783: shouldn't it be `f.get_matrix()` or `f.get_matrix(None)` then?

AhmetNSimsek: Noth works at the end but eventually, it just takes `subject=None`. But perhaps better to use `f.get_matrix()` as the default behaviour might change later.

xgui3783: but you used `f.get_matrix(subject=FileNotFoundError)`

am I missing something?

e2e/features/connectivity/test_connectivity.py
```diff
@@ -22,9 +22,13 @@ def test_connectivity_get_matrix(f: RegionalConnectivity):
     assert isinstance(f, RegionalConnectivity)
     assert len(f.subjects) > 0
     assert all(isinstance(subject, str) for subject in f.subjects)
+    matrix_df = f.get_matrix(subject=FileNotFoundError)
```
xgui3783, line 25:
```suggestion
    matrix_df = f.get_matrix()
```

e2e/features/connectivity/test_connectivity.py
xgui3783: see https://github.com/FZJ-INM1-BDA/siibra-python/pull/347/files#r1172494587

AhmetNSimsek: Oh I thought I did `subject=None`. That is very weird...

AhmetNSimsek: Corrected with [b76161b](https://github.com/FZJ-INM1-BDA/siibra-python/pull/347/commits/b76161b5a5cc3831f2a27b4fa2e8cefebe717240)

AhmetNSimsek:
(The example tests failed due to network error in Allen inst.)

PR #356, opened 2023-04-25
Use unique name for the cache of sparseindices

### LINKED ISSUE #355: Sparseindex naming in cache is not unique
When a `SparseIndex` is created for a `SparseMap`, it creates a name prefix via:
```python
prefix  = f"{self.parcellation.id}_{self.space.id}_{self.maptype}_index"
```
(This is later)However, such a prefix is not unique for parcellations such as Julich Brain 3.0 in which two statistical maps are released.

Suggestion: use `f"{self.name}_index"`, which has to be unique to be in the configuration.

AhmetNSimsek:
Fixes #355

### DIFF
--- a/siibra/volumes/sparsemap.py
+++ b/siibra/volumes/sparsemap.py
@@ -245,7 +245,7 @@ class SparseMap(parcellationmap.Map):
     @property
     def sparse_index(self):
         if self._sparse_index_cached is None:
-            prefix = f"{self.parcellation.id}_{self.space.id}_{self.maptype}_index"
+            prefix = f"{self.name}_index"
             spind = SparseIndex.from_cache(prefix)
             with _volume.SubvolumeProvider.UseCaching():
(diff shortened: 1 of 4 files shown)


siibra/volumes/sparsemap.py
```diff
@@ -245,7 +245,7 @@ def __init__(
     @property
     def sparse_index(self):
         if self._sparse_index_cached is None:
-            prefix = f"{self.parcellation.id}_{self.space.id}_{self.maptype}_index"
+            prefix = f"{self.name}_index"
```
xgui3783, line 248: I would use 

```python
prefix = f"{self.parcellation.id}_{self.space.id}_{self.maptype}_{self.name}_index"
```

lest we accidentally add a parcellationmap with the same name

xgui3783 (commented):
Looks good. 

can you also add e2e to prevent regression?

Issue #334, opened 2023-04-13
Improve sparsemap: Cache sparseindex and update preconfigurations


AhmetNSimsek:
`Sparsemaps` need to be converted to `sparseindex` (represented by 3 files) to be used. It is decided that any supported `sparsemap` now will be precomputed and will be published as a Ebrains dataset. New sparsemaps will go through the same pipeline and will be added to the dataset.

To-do list:
- [x] Update `sparsemap.py` to reflect storing/loading
- [x] `Sparsemap` type will be determined from the preconfiguration.
    - [x] The preconfigurations will be extended with a `sparsemap` field: `is_sparsemap`: bool, `cached`: bool, `url`: str, and `ebrains`: dict. If the map is sparse but not cached, then the current functionality be preserved (i.e., pull the data from individual URLs and create a `sparseindex` on the fly).
    - [x] Allows siibra to decide if a map configuration is sparse or not from the config itself.
    - [x] Update configurations: https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/issues/8
- [x] Write a script to create `sparseindex` for each `sparsemap` and convert all.
- [ ] Create a dataset on Ebrains with the code and the `sparseindex` files. (Each map as a folder with 3 files: probs.txt.gz, bbox.txt.gz, voxel.nii.gz).
- [x] Either: (EDT  `min_num_of_volumes` is reworked)
     - Allow labelled maps with high number of volumes to be represented by sparsemaps or
     - Consider reworking `min_num_of_volumes` for sparsemap representation

AhmetNSimsek:
Relevant branch: [clarifySparsemapUsage](https://github.com/FZJ-INM1-BDA/siibra-python/tree/clarifySparsemapUsage)

AhmetNSimsek:
Since EBRAINS dataset version creation is as time consuming as creating a dataset, for the time being this idea is shelved. Instead of a repo, however, the data will be stored in data-proxy bucket most likely. The key is to handle versioning

PR #363, opened 2023-04-27
Improve sparse map handling

AhmetNSimsek:
This PR introduces:
- Allow a clear definition for a SparseMap from the configuration.
     - Legacy method is preserved
-  Allows fetching SparseIndex of a SparseMap from a precomputed SparseIndex in a zip file.

Connected config branch: https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/tree/clarifySparsemaps

This decreases the required time it takes to fetch any map from SparseMaps as the SparseIndex is no longer required to be calculated on the fly.
No functionality is removed, so if a SparseMaps does not have stored SparseIndex or the zip file is unreachable, then the Index is calculated as before.

See related issue #334

### DIFF
--- a/config_schema/siibra/map/v0.0.1.json
+++ b/config_schema/siibra/map/v0.0.1.json
@@ -11,9 +11,18 @@
             "$ref": "config_schema/at_id.json"
         },
         "sparsemap": {
-            "is_sparsemap": "bool",
-            "cached": "bool",
-            "url": "string"
+            "type": "object",
+            "properties": {
+                "is_sparsemap": {
+                    "type": "boolean"
+                },
+                "cached": {
+                    "type": "boolean"
+                },
+                "url": {
+                    "type": "string"
+                }
+            }
         },
         "volumes": {
             "type": "array",
--- a/config_schema/siibra/map/v0.0.1.json
+++ b/config_schema/siibra/map/v0.0.1.json
@@ -10,6 +10,11 @@
         "parcellation": {
             "$ref": "config_schema/at_id.json"
         },
+        "sparsemap": {
+            "is_sparsemap": "bool",
+            "cached": "bool",
+            "url": "string"
+        },
         "volumes": {
             "type": "array",
             "items": {
--- a/siibra/configuration/factory.py
+++ b/siibra/configuration/factory.py
@@ -269,7 +269,7 @@ class Factory:
         identifier = f"{spec['@type'].replace('/','-')}_{basename}"
         volumes = cls.extract_volumes(spec)
 
-        if ("sparsemap" in spec) and spec.get("sparsemap")["is_sparsemap"]:
+        if ("sparsemap" in spec) and spec.get("sparsemap").get("is_sparsemap"):
             Maptype = sparsemap.SparseMap
             return Maptype(
                 identifier=spec.get("@id", identifier),
@@ -290,7 +290,7 @@ class Factory:
         Maptype = parcellationmap.Map
         if len(volumes) > MIN_VOLUMES_FOR_SPARSE_MAP:
             logger.debug(
-                f"Using sparse map for {spec['filename']} to code its "
+                f"Using sparse map for {spec.get('filename')} to code its "
                 f"{len(volumes)} volumes efficiently."
             )
             Maptype = sparsemap.SparseMap
@@ -302,7 +302,7 @@ class Factory:
             ) + 1
             if max_z > MIN_VOLUMES_FOR_SPARSE_MAP:
                 logger.debug(
-                    f"Using sparse map for {spec['filename']} to code its "
+                    f"Using sparse map for {spec.get('filename')} to code its "
                     f"{max_z} z levels efficiently."
                 )
                 Maptype = sparsemap.SparseMap
--- a/siibra/configuration/factory.py
+++ b/siibra/configuration/factory.py
@@ -269,10 +269,29 @@ class Factory:
         identifier = f"{spec['@type'].replace('/','-')}_{basename}"
         volumes = cls.extract_volumes(spec)
 
+        if ("sparsemap" in spec) and spec.get("sparsemap")["is_sparsemap"]:
+            Maptype = sparsemap.SparseMap
+            return Maptype(
+                identifier=spec.get("@id", identifier),
+                name=spec.get("name", name),
+                space_spec=spec.get("space", {}),
+                parcellation_spec=spec.get("parcellation", {}),
+                indices=spec.get("indices", {}),
+                volumes=volumes,
+                shortname=spec.get("shortName", ""),
+                description=spec.get("description"),
+                modality=spec.get("modality"),
+                publications=spec.get("publications", []),
+                datasets=cls.extract_datasets(spec),
+                is_cached=spec.get("sparsemap").get("cached", False),
+                cache_url=spec.get("sparsemap").get("url", "")
+            )
+
         Maptype = parcellationmap.Map
         if len(volumes) > MIN_VOLUMES_FOR_SPARSE_MAP:
             logger.debug(
-                f"Using sparse map for {spec['filename']} to code its {len(volumes)} volumes efficiently."
+                f"Using sparse map for {spec['filename']} to code its "
+                f"{len(volumes)} volumes efficiently."
             )
             Maptype = sparsemap.SparseMap
         else:
@@ -283,7 +302,8 @@ class Factory:
             ) + 1
             if max_z > MIN_VOLUMES_FOR_SPARSE_MAP:
                 logger.debug(
-                    f"Using sparse map for {spec['filename']} to code its {max_z} z levels efficiently."
+                    f"Using sparse map for {spec['filename']} to code its "
+                    f"{max_z} z levels efficiently."
                 )
                 Maptype = sparsemap.SparseMap
 
--- a/siibra/retrieval/repositories.py
+++ b/siibra/retrieval/repositories.py
@@ -319,6 +319,10 @@ class ZipfileConnector(RepositoryConnector):
     def __eq__(self, other):
         return self.url == other.url
 
+    def clear_cache(self):
+        os.remove(self.zipfile)
+        self._zipfile_cached = None
+
     class FileLoader:
         """
         Loads a file from the zip archive, but mimics the behaviour
@@ -328,8 +332,12 @@ class ZipfileConnector(RepositoryConnector):
             self.zipfile = zipfile
             self.filename = filename
             self.func = decode_func
-            self.cached = True
+            self.cachefile = CACHE.build_filename(zipfile+filename)
 
+        @property
+        def cached(self):
+            return os.path.isfile(self.cachefile)
+        
         @property
         def data(self):
             container = ZipFile(self.zipfile)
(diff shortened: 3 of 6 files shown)


AhmetNSimsek:
Also, added an option to pull zipped sparseIndices from a [gitlab repo](https://jugit.fz-juelich.de/t.dickscheid/siibra-sparse-indices).

siibra/configuration/factory.py
```diff
@@ -269,10 +269,29 @@ def build_map(cls, spec):
         identifier = f"{spec['@type'].replace('/','-')}_{basename}"
         volumes = cls.extract_volumes(spec)
 
+        if ("sparsemap" in spec) and spec.get("sparsemap")["is_sparsemap"]:
```
xgui3783, line 272: I believe map has spec defined in json schema, and sparsemap is not amongst them. 

we should add it if we are to introduce it in siibra-python

also I would use `.get("is_sparsemap")` rather than `["is_sparsemap"]`. Latter can throw IndexError

siibra/configuration/factory.py
```diff
@@ -269,10 +269,29 @@ def build_map(cls, spec):
...
+                datasets=cls.extract_datasets(spec),
+                is_cached=spec.get("sparsemap").get("cached", False),
+                cache_url=spec.get("sparsemap").get("url", "")
+            )
+
         Maptype = parcellationmap.Map
         if len(volumes) > MIN_VOLUMES_FOR_SPARSE_MAP:
             logger.debug(
-                f"Using sparse map for {spec['filename']} to code its {len(volumes)} volumes efficiently."
+                f"Using sparse map for {spec['filename']} to code its "
```
xgui3783, line 293: ditto above re `.get` vs `[key]`

siibra/retrieval/repositories.py
```diff
@@ -319,6 +319,10 @@ def search_files(self, folder="", suffix="", recursive=False):
     def __eq__(self, other):
         return self.url == other.url
 
+    def clear_cache(self):
+        os.remove(self.zipfile)
+        self._zipfile_cached = None
+
```
xgui3783, line 325: I have a feeling that we should let `siibra.cache` to manage the lifecycle of caches, and not manage it on the repository level.

siibra/retrieval/repositories.py
```diff
@@ -319,6 +319,10 @@ def search_files(self, folder="", suffix="", recursive=False):
...
     class FileLoader:
```
xgui3783, line 326: we should consider rename this class `ZipFileLoader`

I was mightly confused what a property called `zipfile` is doing in the class `FileLoader`

siibra/retrieval/requests.py
```diff
@@ -179,6 +179,8 @@ def _retrieve(self, block_size=1024, min_bytesize_with_no_progress_info=2e8):
                     f.write(data)
             if size_bytes > min_bytesize_with_no_progress_info:
                 progress_bar.close()
+            if self.refresh and os.path.isfile(self.cachefile):
+                os.remove(self.cachefile)
```
xgui3783, line 183: is this necessary? I would imagine `os.rename` would overwrite already?

siibra/volumes/sparsemap.py
```diff
@@ -131,7 +133,7 @@ def to_cache(self, prefix: str):
                 )
 
     @classmethod
-    def from_cache(cls, prefix: str):
+    def _from_cache(cls, prefix: str):
```
xgui3783, line 136: we could probably turn this from a classmethod to a staticmethod, since cls variable is not really used

I would argue `prefix` variable should be called `cache_name` or `cache_key`

siibra/volumes/sparsemap.py
```diff
@@ -239,14 +243,33 @@ def __init__(
...
+            if spind is None:
+                logger.debug("Loading SparseIndex from Gitlab.")
+                gconn = GitlabConnector(self._GITLAB_SERVER, self._GITLAB_PROJECT, "main")
+                zip_fname = f"{self.name.replace(' ', '_')}_index.zip"
+                try:
+                    assert zip_fname in gconn.search_files(), f"{zip_fname} is not in {gconn}."
+                    zipfile = gconn.get_loader(zip_fname).url
+                    spind = self.load_zipped_sparseindex(zipfile)
+                except Exception as e:
+                    logger.debug("Could not load SparseIndex from Gitlab:\n", exc_info=1)
```
xgui3783, line 272: Correct me if I am wrong. The logic here is:

1/ check if sparseindex is cached locally. if not...
2/ check for sparseindex cache on gitlab, if does not exist ...
3/ compute sparse index from scratch

I think the logic can be simpler (or at least more maintainable) if we abstract the fetching. 

So that, in future, if we have 15 different ways to fetch sparse index, we don't have a 1000 line method property.

I created something similar for [http request](https://github.com/FZJ-INM1-BDA/siibra-python/blob/edf8c20/siibra/retrieval/requests.py#L534-L552) earlier, but can be adapted to this usecase.

siibra/volumes/sparsemap.py
```diff
@@ -275,6 +298,70 @@ def affine(self):
...
+                for suffix in suffices:
+                    zipf.write(
+                        filename=cache.CACHE.build_filename(self._cache_prefix, suffix),
+                        arcname=path.basename(f"{filename}{suffix}"),
+                        compress_type=ZIP_DEFLATED
+                    )
+        except Exception as e:
+            logger.error("Could not save SparseIndex:\n")
+            raise e
+        logger.info("SparseIndex is saved.")
```
xgui3783, line 333: I can't figure out how the data is written. filename and arcname is provided to zipf, but... what about data?

xgui3783 (commented):
I would like to hear your opinion on the MultiSourcedCache, in particular. I think it can make the sparse_index property code a lot cleaner

siibra/retrieval/repositories.py
AhmetNSimsek: That would be a lot cleaner indeed

siibra/retrieval/repositories.py
AhmetNSimsek: I am glad I was not the only one 😅
I will update it then

siibra/retrieval/requests.py
AhmetNSimsek: It unfortunately does not because it fails saying the file already exists.

siibra/volumes/sparsemap.py
AhmetNSimsek: Both fair points

siibra/retrieval/requests.py
xgui3783: https://docs.python.org/3/library/os.html#os.rename

seems like a windows thing

jokes aside, we should consider use `shutil` rather than `os`

siibra/volumes/sparsemap.py
AhmetNSimsek: Ah almost but not quite.
0/ check if sparseindex cached (line: 256)
1/ check if there is a a precomputed sparseindex if _sparseindex_zip_url is taken from the json. (goal: ebrains dataset)
2/ check gitlab cache
3/ compute from scratch

I see your point. That could indeed be useful. Original idea was to use the ebrains dataset only but there are a few things to be approved before it is done. So I decided to implement gitlab as well.
I will take a look at the multisourcerequest quickly

siibra/retrieval/requests.py
AhmetNSimsek: Oh yeah quite possible...
Hmm, perhaps. This would be a bigger overhaul tho, no? Because we use os everywhere as far as I can remember.

siibra/volumes/sparsemap.py
AhmetNSimsek: Ah I wanted to avoid the actual writing process as SparseIndex files will already be in the cache (or will be calculated on the fly to be in the cache). It just basically writes in a zip file from the siibra cache. So then no need to load the data etc or write as `_to_cache` does.

AhmetNSimsek: I introduced `_cache_prefix` property for this reason; if that is changed, everything related to the local cache will change along with it. So we just need to update the property later, no?

siibra/volumes/sparsemap.py
AhmetNSimsek: Could you take a look at this commit? b05b927e7aaeb5931dbd701e45c7deee785b7968

siibra/retrieval/requests.py
AhmetNSimsek: https://docs.python.org/3/library/os.html#os.replace
We can use `os.replace` instead as it won't throw FileExistsError.

AhmetNSimsek: Now, it throws FileNotFoundError from test_request.py (`test_file_lock_` and `test_httprequests_retrieve`)

config_schema/siibra/map/v0.0.1.json
```diff
@@ -10,6 +10,11 @@
         "parcellation": {
             "$ref": "config_schema/at_id.json"
         },
+        "sparsemap": {
+            "is_sparsemap": "bool",
+            "cached": "bool",
+            "url": "string"
+        },
```
xgui3783, line 17: this is not valid json schema, IIRC

siibra/volumes/sparsemap.py
xgui3783: But there is nothing inherently unique about the cache name, which, IMO makes this approach flaky. For example, one may update

https://github.com/FZJ-INM1-BDA/siibra-python/blob/73498b5/siibra/volumes/sparsemap.py#L119-L121

and 

https://github.com/FZJ-INM1-BDA/siibra-python/blob/73498b5/siibra/volumes/sparsemap.py#L152-L154

in tandem, but certainly would not expect the filename is used at https://github.com/FZJ-INM1-BDA/siibra-python/blob/73498b5/siibra/volumes/sparsemap.py#L325-L329 

, nor should we expect them to.

xgui3783: my proposal is to have `save_sparseindex` (shouldn't this be a method of sparseindex, rather than sparsemap?) return a tuple, and `load_sparseindex` accept a tuple

This way, the input and expectation of these methods are clear, and the method is pure (do not induce side effects)

siibra/retrieval/requests.py
xgui3783: > Oh yeah quite possible... Hmm, perhaps. This would be a bigger overhaul tho, no? Because we use os everywhere as far as I can remember.

we don't have to change everything all at once. just the codes we are touching

siibra/volumes/sparsemap.py
xgui3783: Hmm, when I proposed multisource approach, I was expecting the sources to be declared, rather than compiled at runtime. 

i.e.

```python
class SparseMap:
    sources = [
        LocalCache(),
        GitCache(), 
    ]
```

then, we could just

```python
for src in sources:
   try:
      data = src.get()
      break
   except:
      pass
else:
   raise NotFoundError
# do something to data
```

this way, the `sparse_index` property does not need to be privy how the sources are implemented, so long it has a `get()` method

AhmetNSimsek:
Note: Based on discussions with @xgui3783, a secondary [branch](https://github.com/FZJ-INM1-BDA/siibra-python/tree/wip_clarifySparsemap) is being developed to further clarify and clean up sparsemap.py

AhmetNSimsek:
Merging this branch for now as it is urgent but later sparsemap.py will be updated with https://github.com/FZJ-INM1-BDA/siibra-python/tree/wip_clarifySparsemap based on the discussion here.

Issue #329, opened 2023-04-04
[enhancement] Maps should have an InstanceTable as well


AhmetNSimsek:
Instance tables provide a great way to check the existing configuration concepts. However, for maps, this has been lacking due to their inherit complexity. @dickscheid and I had discussed this before and had considered an upgrade to the instance tables by making them a data frame that stores extra information such as species. The branch [attrDFforMaps](https://github.com/FZJ-INM1-BDA/siibra-python/tree/attrDFforMaps) introduces such a data frame to maps while instance tables are still used for the core concepts. This branch is made to facilitate the discussion:
- Should all the InstanceTables be replaced by such a data frame? (In such a case, the naming InstanceTable will be preserved.)
- What else can we improve for the maps InstanceDF?
- Should we change the naming of the statistical map files in the configuration? (continuous<->statistical)
- Thoughts about a similar table for features?

Code example:
```python
import siibra
print(siibra.maps.head())
print(siibra.maps.columns)
print(siibra.maps.index)
siibra.maps.query("space == 'MNI Colin 27'")
```

PS: The original code is authored by @dickscheid.

xgui3783:
Just occurred to me one thing:

> Should all the InstanceTables be replaced by such a data frame? (In such a case, the naming InstanceTable will be preserved.)

I would hate to lose static typing offered by instance table.

dickscheid:
Dataframes are very nice in notebooks to plot the instances with some attributes, but in line with Xiao's comment I think we need to clarify a few important functionalities:

* can we do static typing with this?
* can we do autocompletion on the instance keys a suitable interpreter?

Maybe the solution is to provide a data frame listing optionally, using a dedicated function. Then it could be part of the the atlas concept parent class, and we could keep the instance table.

AhmetNSimsek:
-If we enable static typing, the default `__iter__` has to be replaced. Which then in turn breaks the default representation of dataframes. I am thinking about this currently.

- Autocompletion is currently supported
![image](https://user-images.githubusercontent.com/69900321/237055127-916591c4-0024-47f9-9e0e-168656b950ad.png)

> Maybe the solution is to provide a data frame listing optionally, using a dedicated function. Then it could be part of the the atlas concept parent class, and we could keep the instance table.

I think this might work. I will give it a try and update here

xgui3783:
[...]

PR #364, opened 2023-04-28
InstanceDataframe for maps (strictly additive, does not remove InstanceTable)

AhmetNSimsek:
This PR adds an instance table in the form of a data frame for maps (#329). It is strictly additive and does not remove any existing functionality for the other instance tables of class `InstanceTable`.
While more discussion and additions could be had, this is a highly requested feature. Therefore, IMO we should add it as is and continue discussing the remainder under a new issue.
(`InstanceDataframe` could be expanded with similar functionalities of `InstanceTable`, however, for the time bein the goal is to provide an overview. Also, `query` provides a lot of functionalites.)
Usage:
```python
import siibra
print(siibra.maps.head())
print(siibra.maps.columns)
print(siibra.maps.index)
siibra.maps.query("space == 'MNI Colin 27'")
```

### DIFF
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -231,39 +231,8 @@ class InstanceTable(Generic[T], Iterable):
                             for attribute in ['parcellation', 'space', 'maptype']
                         }
                     )
-            self._dataframe_cached = self.InstanceDataframe(self._elements, attrs)
+            self._dataframe_cached = pd.DataFrame(index=list(self._elements.keys()), data=attrs)
         return self._dataframe_cached
-
-    class InstanceDataframe(pd.DataFrame):
-
-        def __init__(self, objs: dict, attrs: list):
-            pd.DataFrame.__init__(self, data=attrs, index=list(objs.keys()))
-            import warnings
-            with warnings.catch_warnings():
-                warnings.filterwarnings("ignore", category=UserWarning)
-                self._objs = objs
-
-        def __dir__(self):
-            # make autocompletion work on the row names
-            return list(self.index)
-
-        def __getattr__(self, attr):
-            # make the row name return the corresponding object
-            if attr in self.index:
-                return self._objs[attr]
-            return super().__getattr__(attr)
-
-        def query(self, expr: str):
-            # allow to build a filtered instance table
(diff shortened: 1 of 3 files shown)


AhmetNSimsek:
Note: iterator of dataframes returns headers hence loops utilizing the map registry do not work. A simple iterator of the form
```python
def __iter__(self) -> Iterator[T]:
        yield from self._objs.values()
```
enables this. However, then it breaks the default representation of `pandas.Dataframes`.

siibra/commons.py
```diff
@@ -215,6 +217,53 @@ def __getattr__(self, index) -> T:
...
+            # allow to build a filtered instance table
+            df = pd.DataFrame.query(self, expr, inplace=False)
+            return self.__class__(
+                objs={
+                    k: v for k, v in self._objs.items()
+                    if k in df.index
+                },
+                attrs=list(dict(df.T).values())
+            )
+
```
xgui3783, line 266: What's the benefit of having a bespoke class for the dataframe? couldn't `.dataframe` just return a dataframe?

siibra/commons.py
```diff
@@ -215,6 +217,53 @@ def __getattr__(self, index) -> T:
...
+            for i, val in enumerate(values):
+                attrs.append({'name': val.name, 'species': str(val.species)})
+                if hasattr(val, 'maptype'):
+                    attrs[i].update(
+                        {
+                            attribute: val.__getattribute__(attribute).name
+                            for attribute in ['parcellation', 'space', 'maptype']
+                        }
+                    )
+            self._dataframe_cached = self.InstanceDataframe(self._elements, attrs)
```
xgui3783, line 234: (cont'ed)

so we would have:

```python
self._dataframe_cached = pd.DataFrame(index=self._elements.keys(), data=attrs)
```

here

siibra/commons.py
AhmetNSimsek: I wanted to preserve some extra functionality like autocomplete but in hind sight, I see that one can just use the standard registry for such functionality.

xgui3783 (approved):
LGTM

Issue #332, opened 2023-04-12
JBA 3.0 labelled maps have regions that has no apperant match in the parcellation


AhmetNSimsek:
as of 935503c7314170cacd21b2d0b8fcbd5a2c3555f3

This is not a bug but a data issue.

```python
import siibra
p = siibra.parcellations["julich 3.0"]
mp = siibra.get_map(p, "colin 27")  # or mni152
for r in mp.regions:
    try:
        p.get_region(r)
    except:
        print(r)
```
Expected to be empty but the output:
```
Medial Accumbens, ventral Striatum (Basal Ganglia) left
Medial Accumbens, ventral Striatum (Basal Ganglia) right
Fundus of Caudate Nucleus, ventral Striatum (Basal Ganglia) left
Fundus of Caudate Nucleus, ventral Striatum (Basal Ganglia) right
Fundus of Putamen, Ventral Striatum (Basal Ganglia) left
Fundus of Putamen, Ventral Striatum (Basal Ganglia) right
Lateral Accumbens, ventral Striatum (Basal Ganglia) left
Lateral Accumbens, ventral Striatum (Basal Ganglia) right
VP, ventral Pallidum (Basal Ganglia) left
VP, ventral Pallidum (Basal Ganglia) right
```

AhmetNSimsek:
Fixed by https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/new?merge_request%5Bsource_project_id%5D=3484&merge_request%5Bsource_branch%5D=jba3labmaps_fixMismatchingNames&merge_request%5Btarget_project_id%5D=3484&merge_request%5Btarget_branch%5D=master

AhmetNSimsek:
TODO: add test

Issue #291, opened 2023-03-02
[bug] some region mapped by big brain julich brain 2.9 cannot be found on parcellation


xgui3783:
region listed below:

```
MGBd (Metathalamus) left 
MGBd (Metathalamus) right 
MGBm (Metathalamus) left 
MGBm (Metathalamus) right 
MGBv (Metathalamus) left 
MGBv (Metathalamus) right 
CGL.lam1 (Metathalamus) left 
CGL.lam1 (Metathalamus) right 
CGL.lam2 (Metathalamus) left 
CGL.lam2 (Metathalamus) right 
CGL.lam3 (Metathalamus) left 
CGL.lam3 (Metathalamus) right 
CGL.lam4 (Metathalamus) left 
CGL.lam4 (Metathalamus) right 
CGL.lam5 (Metathalamus) left 
CGL.lam5 (Metathalamus) right 
CGL.lam6 (Metathalamus) left 
CGL.lam6 (Metathalamus) right 
```

are mapped in julich brain 2.9 in big brain, but does not appear on parcellation region hierarchy

This is a problem because

a/ there exist links in KG that point to viewer link
b/ since siibra-python do not recognise these regions exist in julich brain 2.9, they do not appear in region hierarchy, and no features will be shown

SBludau:
I think this is due to the fact that some of the maps come from different sources.  All probability maps are based on 10 brains and are drawn in the OnlineSectionTracer and calculated by Hartmut. Hartmut generates his terminology file based on his database and the OnlineSectionTracer.

Special BigBrain mappings like the CGL lamina (e.g. CGL.lam1 (metathalamus) left) are calculated by Christian.

There must be a problem with the coordination and linking of the different datasets.

AhmetNSimsek:
https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/82 solves the issue

AhmetNSimsek:
Solved with https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/42

PR #373, opened 2023-05-05
Test configured maps to get regions from parcellations

AhmetNSimsek:
The reasons for the test:

#332 and https://github.com/FZJ-INM1-BDA/siibra-python/issues/291

Additionally:
- monkey mebrains labelled
   - Pallidum
- fsaverage jba30 labelled
   - BrainStemMask left
   - BrainStemMask right


AhmetNSimsek:
Note: The failing tests are exactly the reason this test is added for. Should be in main.

xgui3783:
proposed remedy:

either:
- update appropriate tag siibra configuration and retrun test. or...
- create new tag on siibra configuration and bump version in this PR

AhmetNSimsek:
Sebastian informed me that `BrainStemMask` cannot be a part of the region tree. But it was required for the creation of fsaverage surface map.
Options I see:
- `BrainStemMask` does not need to be in the map configuration.
- or we can add an exception case for the test
What do you think @dickscheid?

xgui3783 (approved):
looks good. but I would add CI test to brainscapes configuration, which makes more sense than e2e here.

AhmetNSimsek:
> looks good. but I would add CI test to brainscapes configuration, which makes more sense than e2e here.

I agree that the proper place is brainscapes-configurations. But I think we need `siibra` for this check since the region names are not exact matches and there could be subset matches etc. `get_region` simplifies life a lot. We could import siibra in _ci to check... Thoughts?

xgui3783:
> > looks good. but I would add CI test to brainscapes configuration, which makes more sense than e2e here.
> 
> I agree that the proper place is brainscapes-configurations. But I think we need `siibra` for this check since the region names are not exact matches and there could be subset matches etc. `get_region` simplifies life a lot. We could import siibra in _ci to check... Thoughts?

the keys in indicies *should* be an exact match to that of parcellation, no?

xgui3783:
All the reason to have the CI on config, so that it can be caught as early
as possible?

AhmetNSimsek:
What do you think of the following:
https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/commit/bcf09c2c1f1414d753a47138640c8b7901b58a55

AhmetNSimsek:
Requires https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/merge_requests/87

AhmetNSimsek:
TODO: add a field to maps for regions that are mapped but not represented in the parcellation, such as brain stem mask in fsaverage julich brain 3.

Issue #338, opened 2023-04-18
Region assignment: `None` in `containedness` despite point uncertainty


JThoennissen:
I’m assigning coordinates to regions with an uncertainty (exactly like in the [documentation](https://siibra-python.readthedocs.io/en/latest/examples/05_anatomical_assignment/001_coordinates.html)). In my code as well as in the documentation the values for `containedness` are `None`,  which shouldn’t be the case when specifying uncertainty. When running corresponding code on v0.3a27, I get values for the `Contained` key.

**Python version**: 3.8.10
**siibra version**: 0.4a46
**Code**:
```python
import siibra

with siibra.QUIET:  # suppress progress output
    julich_pmaps = siibra.get_map(
        parcellation="julich 2.9",
        space="mni152",
        maptype="statistical"
    )

point_uncertain = siibra.Point((27.75, -32.0, 63.725), space='mni152', sigma_mm=3.)
with siibra.QUIET:  # suppress progress output
    assignments = julich_pmaps.assign(point_uncertain)

print(assignments)
```
**Result**:
![image](https://user-images.githubusercontent.com/32842931/232752982-88cb1f18-e513-4973-b0d6-17057f290efd.png)

AhmetNSimsek:
As of 8da09309b379ea50b7f183437094f2c6dc424432, this remains to be a bug.
To-Do:
- [ ] Find the bug source
- [ ] Test based on https://siibra-python.readthedocs.io/en/latest/examples/05_anatomical_assignment/001_coordinates.html (The current outputs are misleading. The last cell should produce a plot)
- [ ] Update the e2e test to consider None to avoid false positive

AhmetNSimsek:
The issue seems to be resolved by no other commit on 28.04.2023. Could be caused by a network issue?

JThoennissen:
It seems to be working for me as well 👍

AhmetNSimsek:
The issue reoccurs randomly. TODO:
- [ ] Write a mock network test

AhmetNSimsek:
Note:
`point_uncertain = siibra.Point((27.75, -32.0, 63.725), space='mni152', sigma_mm=3.)`
should produce a full set of values but
`point = siibra.Point((27.75, -32.0, 63.725), space='mni152')`
will only produce not `Nan` for "map value"

PR #378, opened 2023-05-11
Point assignments also have map value

AhmetNSimsek:
During implementation of dataclass for assignments, the 'map value' was forgotten to be assigned for points. Fixes
https://github.com/FZJ-INM1-BDA/siibra-python/issues/338


xgui3783:
can you add some e2e so regression doesn't occur in the future?

PR #381, opened 2023-05-16
Detect csv delimiter with Python’s builtin sniffer tool

AhmetNSimsek:
The use of ";" as the default csv decoder delimiter appears to be originally written to parse 1000 brains data. However, now this is resolved in `decoder` section in the corresponding json file. It is inherently not intuitive to use ";" as the default of the csv decoder.

Before merging, the documentation should be checked for any issues with examples or data ingestion in general.


xgui3783 (approved):
LGTM

Issue #386, opened 2023-05-22
[bug] mni152 nii template returns None


xgui3783:
as of 5a649faf94f07abcd91c0c8979a58bcd07b9c7bf

code to reproduce:

```python
import siibra
space = siibra.spaces["mni152"]
assert space.get_template().fetch("nii") 
```

expected result, assert succeeds

actual behavior: assert fails (assert None)

prime cause: mni152 providers are `zip/nii` and `neuroglancer/precomputed`. Current implementation uses exact format match, as such `nii` does not match `zip/nii`

proposed solution: curate a dict of aliases. e.g. when user query nii, also queries zip/nii

AhmetNSimsek:
I do not agree that this is a bug since `space.get_template().formats` does not list `nii` alone. But I agree that asking for `nii` should be able to call `zip/nii` as well.

I agree with the dictionary idea. We could also add `nifti` as another alias. This could be expanded for others later as well. (potentially we could expand 'image' and 'mesh' with this system as well)

Issue #318, opened 2023-03-30
[bug] attempt to fetch volumes with multiple sources can occassionally result in undeterministic behaviour


xgui3783:
as of 0.4a35

When volume with multiple providers (e.g. nii + neuroglancer/precomputed) calls fetch, network conditions will determine which provider is used. 

e.g. for [mni152 jb29 labelled](https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/blob/siibra-0.4a35/maps/mni152-jba29-labelled.json#L11-19) where both nii and precomputed are provided, nii will be attempted first (per https://github.com/FZJ-INM1-BDA/siibra-python/blob/01d3d09/siibra/volumes/volume.py#L40-L44 ). If it fails (src is unreachable, in maintenance, being rate limited), neuroglancer precomputed source will be used. 

precomputed provider fetch seems to default to lowest resolution.

As a result, voxel count will likely be a factor of 8 less.

code to reproduce:

caveat:
- need to install test dep `requests-mock`
- needs to run `siibra.cache.clear()` between each run

```python
import siibra
import numpy as np
import requests_mock

siibra.cache.clear()

url = "https://neuroglancer.humanbrainproject.eu/precomputed/data-repo-ng-bot/20210714-julichbrain_v290_correctaffine/hbp-d000007_julichbrain-cytoatlas_dev/MPMs/29.0/GapMapPublicMPMAtlas_l_N10_nlin2StdICBM152asym2009c_29_publicDOI_b1bcc0b5a127f5a917f0d8b0e869be5d.nii.gz"

def main():
    region = siibra.get_region("2.9", "area 1 (postcg) left")
    rmap = region.fetch_regional_map("mni152")
    print(np.count_nonzero(rmap.get_fdata()))

def mocked():
    with requests_mock.Mocker(real_http=True) as mocker:
        mocker.register_uri(url=url, method="get", status_code=404, text="oh no")
        main()

main() # prints 6203

```

clear siibra cache

```python
import siibra
import numpy as np
import requests_mock

siibra.cache.clear()

[...]

PR #391, opened 2023-05-31
If no format is given, only try to fetch single format else fail

### LINKED ISSUE #368: [bug] labelled map and/or mask and/or statistic map inconsistent result
as of [v0.4a47](https://github.com/FZJ-INM1-BDA/siibra-python/tree/v0.4a47)

min code to reproduce:

```python

MNI152_ID="minds/core/referencespace/v1.0.0/dafcffc5-4826-4bf1-8ff6-46b8a31ff8e2"
COLIN_ID="minds/core/referencespace/v1.0.0/7f39f7be-445b-47c0-9791-e971c0b6d992"

JBA_29_ID="minds/core/parcellationatlas/v1.0.0/94c1125b-b87e-45e4-901c-00daee7f2579-290"
JBA_30_ID="minds/core/parcellationatlas/v1.0.0/94c1125b-b87e-45e4-901c-00daee7f2579-300"

HOC1_RIGHT="Area hOc1 (V1, 17, CalcS) - right hemisphere"
FP1_RIGHT="Area Fp1 (FPole) - right hemisphere"

STATISTIC_ENDPOINT="statistical"
LABELLED_ENDPOINT="labelled"

map_shape_args = product(
    ((MNI152_ID, (193, 229, 193)),),
    (JBA_29_ID,),
    (STATISTIC_ENDPOINT, LABELLED_ENDPOINT),
    (HOC1_RIGHT, FP1_RIGHT, None),
)

@pytest.mark.parametrize('space_shape,parc_id,map_endpoint,region_name', map_shape_args)
def test_map_shape(space_shape,parc_id,map_endpoint,region_name):
    if region_name is None and map_endpoint == STATISTIC_ENDPOINT:
        assert True
        return
    space_id, expected_shape = space_shape
    
    volume_data = None
    if region_name is not None:
        region = siibra.get_region(parc_id, region_name)
        volume_data = region.fetch_regional_map(space_id, map_endpoint)
    else:
        labelled_map = siibra.get_map(parc_id, space_id, map_endpoint)
        assert labelled_map is not None
        volume_data = labelled_map.fetch()
    
    assert volume_data
    assert volume_data.get_fdata().shape == expected_shape, f"{volume_data.get_fdata().shape}, {expected_shape}, {region_name}, {map_endpoint}, {space_id}"

```

You may have to run a few times to see the assertion error. see https://github.com/FZJ-INM1-BDA/siibra-python/actions/runs/4871591123

particularly e2e succeeding in py3.9 but failing in py3.10

expected: returned nifti shape to be consistent

maybe related to https://github.com/FZJ-INM1-BDA/siibra-python/issues/318

AhmetNSimsek:
Provides a solution to #318 (and closes #368):
- Fetch the first supported format matching the volume's providers unless the user specifies the volume `format`.
- Try the selected format 6 times and fail.
- On fail, inform the user of the exception, the autoselected format, and other available formats.

Also, adds a lookup dictionary for format shorthand (solves #386)

### DIFF
--- a/siibra/volumes/volume.py
+++ b/siibra/volumes/volume.py
@@ -22,6 +22,7 @@ import nibabel as nib
 from abc import ABC, abstractmethod
 from typing import List, Dict, Union, Set, TYPE_CHECKING
 import json
+from time import sleep
 
 if TYPE_CHECKING:
     from ..retrieval.datasets import EbrainsDataset
@@ -57,15 +58,7 @@ class Volume:
         "mesh": MESH_FORMATS,
         "surface": MESH_FORMATS,
         "nifti": ["nii", "zip/nii"],
-        "nii": ["nii", "zip/nii"],
-        "neuroglancer": [
-            "neuroglancer/precomputed",
-            "neuroglancer/precompmesh",
-            "neuroglancer/precompmesh/surface"],
-        "neuroglancer mesh": [
-            "neuroglancer/precompmesh",
-            "neuroglancer/precompmesh/surface"
-        ]
+        "nii": ["nii", "zip/nii"]
     }
 
     def __init__(
@@ -192,7 +185,7 @@ class Volume:
 
         if format is None:
             requested_formats = self.SUPPORTED_FORMATS
-        elif format in self._FORMAT_LOOKUP.keys():  # allow use of aliases
+        elif format in self._FORMAT_LOOKUP:  # allow use of aliases
             requested_formats = self._FORMAT_LOOKUP[format]
         elif format in self.SUPPORTED_FORMATS:
             requested_formats = [format]
@@ -205,6 +198,7 @@ class Volume:
                 selected_format = fmt
                 break
         else:
+            logger.debug(f"Requested format was {format}, selected format is {selected_format}")
             raise ValueError(f"Invalid format requested: {format}")
         
         # try the selected format only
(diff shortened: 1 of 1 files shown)


AhmetNSimsek:
Examples are failing due to Allen institute (on main as well). Rerunning now but the PR is ready otherwise

siibra/volumes/volume.py
```diff
@@ -175,29 +192,38 @@ def fetch(
...
-                    continue
+                selected_format = fmt
+                break
+        else:
+            raise ValueError(f"Invalid format requested: {format}")
+        
+        # try the selected format only
+        for try_count in range(6):
+            try:
+                if selected_format == "gii-label":
```
xgui3783, line 213: This is still oh so very ugly =(

siibra/volumes/volume.py
```diff
@@ -175,29 +192,38 @@ def fetch(
 
         if format is None:
             requested_formats = self.SUPPORTED_FORMATS
-        elif format == 'mesh':
-            requested_formats = self.MESH_FORMATS
-        elif format == 'image':
-            requested_formats = self.IMAGE_FORMATS
+        elif format in self._FORMAT_LOOKUP.keys():  # allow use of aliases
```
xgui3783, line 195: I *think* the `.key()` call is not needed. `val in dictionary` already checks if `val` is a key of `dictionary`

siibra/volumes/volume.py
```diff
@@ -175,29 +192,38 @@ def fetch(
...
+            requested_formats = self._FORMAT_LOOKUP[format]
         elif format in self.SUPPORTED_FORMATS:
             requested_formats = [format]
```
xgui3783, line 198: I would reverse the order. i.e. if format *exactly* matches an entry in `SUPPORTED_FORMATS`, then it will ignore the alias.

siibra/volumes/volume.py
```diff
@@ -52,6 +52,22 @@ class Volume:
...
+        "nifti": ["nii", "zip/nii"],
+        "nii": ["nii", "zip/nii"],
+        "neuroglancer": [
+            "neuroglancer/precomputed",
+            "neuroglancer/precompmesh",
+            "neuroglancer/precompmesh/surface"],
+        "neuroglancer mesh": [
+            "neuroglancer/precompmesh",
+            "neuroglancer/precompmesh/surface"
+        ]
```
xgui3783, line 68: I wouldn't add these aliases.

xgui3783 (commented):
minor changes, otherwise, looks solid. 

please also add tests

siibra/volumes/volume.py
AhmetNSimsek: Initially, that is how I thought as well. However, when user enters 'nii' it accepts it in `requested_formats` and then can fail if 'nii' is not in `self.formats' but 'zip/nii' is. Let me see if I can do another rework

xgui3783: hmm good point

Issue #292, opened 2023-03-06
Feature configuration folder system does not work when using local configuration on Windows


AhmetNSimsek:
When one uses local configuration folders on Windows, the configuration folders take the shape 'features\\connectivity\\regional\\streamlinelengths' instead of 'features/connectivity/regional/streamlinelengths'. This breaks the usage of a local configuration of the feature conf_folder system. Note that this is not an issue for atlases, maps, etc. as they are shipped in single folder.

xgui3783:
the following needs to be retrofitted using os independent:

- [ ] feature boostrap folder
- [ ] accessing ebrains snapshot

AhmetNSimsek:
solved with #482

PR #394, opened 2023-06-12
Allow using local configs with feature folder hierarchy on Windows

AhmetNSimsek:
Currently, when local configuration is used on Windows machines, the feature folder system cannot be used. This PR addresses this issue.
Related to #292. (we can use the same solution for ebrains snapshot)

### DIFF
--- a/siibra/features/feature.py
+++ b/siibra/features/feature.py
@@ -15,7 +15,7 @@
 
 from . import anchor as _anchor
 
-from ..commons import logger, InstanceTable, siibra_tqdm
+from ..commons import logger, InstanceTable, siibra_tqdm, normalize_path
 from ..core import concept
 from ..core import space, region, parcellation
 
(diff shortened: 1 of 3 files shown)


siibra/features/feature.py
```diff
@@ -144,7 +144,11 @@ def get_instances(cls, **kwargs) -> List['Feature']:
         from ..configuration.configuration import Configuration
         conf = Configuration()
         Configuration.register_cleanup(cls.clean_instances)
-        assert cls._configuration_folder in conf.folders
+        try:
+            assert cls._configuration_folder in conf.folders
+        except Exception:
+            logger.debug("Configuration folder assertion has failed:", exc_info=1)
+            assert normalize_path(cls._configuration_folder) in conf.folders, "Configuration folder assertion has failed"
```
xgui3783, line 151:
```suggestion
        try:
            assert normalize_path(cls._configuration_folder) in conf.folders, "Configuration folder assertion has failed"
        except Exception:
            logger.debug("Configuration folder assertion has failed:", exc_info=1)
            raise
```

AhmetNSimsek: I had initally written like this but unfortunately this fails when using the online repo on Windows because the path is normalized according to Windows. Any ideas on changing `normalize_path` so that this approach works?

xgui3783: good point.

Does this part need to changed at all (from the current main)?

since the path gets normalized during `conf.build_objects` later any way?

AhmetNSimsek: We can skip this assertion completely in principle and then `build_objects ` will warn the user instead of raising an exception. But to keep the check, we need this here too.

AhmetNSimsek: What do you think @xgui3783?

AhmetNSimsek:
Closing in favor of https://github.com/FZJ-INM1-BDA/siibra-python/pull/421 which will provide a more generalized solution.

PR #397, opened 2023-06-20
Allow digesting Ebrains Dataset V3 information

AhmetNSimsek:
```python
from siibra.retrieval.datasets import EbrainsV3Dataset
dataset = EbrainsV3Dataset("5a16d948-8d1c-400c-b797-8a7ad29944b2")
print(dataset.version_ids)  # useful for extracting a specific version
print(dataset.contributors)  # a list of contributors collated from versions of the dataset
```


AhmetNSimsek:
closes https://github.com/FZJ-INM1-BDA/siibra-python/issues/243

Issue #377, opened 2023-05-10
BigBrain intensity profiles is difficult to handle as type corticalprofile


AhmetNSimsek:
As of 0eaccb42f5888f076f5af8ea247410d1141276aa, when BigBrain intensity profiles is requested, the results are returned as several hundred dataframes with predefined indices (the same for all) and the same header. Each of these feature objects has their own distinct point coordinate where they are anchored to.
```python
import siibra
fts = siibra.features.get(siibra.get_region("julich 2.9", "hoc1"), siibra.features.cellular.BigBrainIntensityProfile)
print([f.location.coordinate for f in fts][0:10])
print(len(fts))
```

It was discussed that the results of such a live query should be aggregated into one DataFrame. The following could be one possibility for end result:
```python
              x          y          z      0.0    0.005     0.01  ...     0.97    0.975     0.98    0.985     0.99    0.995
0    -40.117901 -31.197701  20.138700   7428.0   7511.0   7651.0  ...  17971.0  18520.0  19045.0  19580.0  19951.0  20175.0     
1    -22.734900  60.743099  18.922501   1971.0   2235.0   2781.0  ...  14433.0  15358.0  16213.0  17016.0  17574.0  17898.0     
2    -15.442300   5.535970  34.340199   4470.0   4682.0   5337.0  ...  15715.0  15530.0  15389.0  15286.0  15199.0  15170.0     
3    -32.547298 -24.788200  28.961700   3307.0   3703.0   4412.0  ...  19223.0  18732.0  18457.0  18119.0  17986.0  17865.0     
4    -27.876200  67.004303  15.580500      0.0      0.0      0.0  ...  12691.0  12393.0  12091.0  11788.0  11604.0  11474.0     
...         ...        ...        ...      ...      ...      ...  ...      ...      ...      ...      ...      ...      ...
```

Also, it might be prudent to change the base class of the bigbrain intensity profiles altogether. This is a bigger design discussion and @dickscheid requested to discuss this here with @xgui3783 and I. Potentially, we could create a spatial feature to use as base class. 

To obtain the above result from current design
```python
import pandas as pd
import numpy as np
cols = ["x","y","z"]
cols.extend(fts[0].data.index.to_list())
df_list = [
    pd.DataFrame(
        np.concatenate(
            (np.asanyarray(f.anchor.location.coordinate).reshape(1,3), f.data.transpose().to_numpy()), axis=1
        ),
        columns=cols
    )
    for f in fts
[...]

Issue #295, opened 2023-03-09
feature.get BigBrainIntensityProfile is non deterministic


xgui3783:
as of 5ae4711de4966fbc1a6e7170ed919e8acf0c3a6e

code to reproduce

```python
import siibra
r = siibra.get_region("2.9", "hoc1")
with siibra.QUIET:
    name0 = siibra.features.get(r, "BigBrainIntensityProfile")[0].name
with siibra.QUIET:
    name1 = siibra.features.get(r, "BigBrainIntensityProfile")[0].name

assert name0 == name1 # assertion fails
```

AhmetNSimsek:
Note: while the names/order are not deterministic, the values are:
```
import siibra
r = siibra.get_region("2.9", "hoc1")
with siibra.QUIET:
    fts0= siibra.features.get(r, "BigBrainIntensityProfile")
with siibra.QUIET:
    fts1 = siibra.features.get(r, "BigBrainIntensityProfile")
c0 = [f.anchor.location.coordinate for f in fts0]
c1 = [f.anchor.location.coordinate for f in fts1]
assert len(c0) == len(c1)
assert len([c for c in c0 if c not in c1])  == 0 and len([c for c in c1 if c not in c0]) == 0
```
This implies that `WagstylProfileLoader.match` returns profiles in non-deterministic order.

AhmetNSimsek:
The source of the issue is found to be the method `feature.match` uses `set` to remove duplicate query results. `set` breaks the ordering. There are alternative solutions such as using a for loop or `dict.fromkeys()`. The former is quite slow. The former is much faster although not as fast as `set`.

["dicts are insertion ordered, dict.fromkeys doesn't lose ordering"](https://stackoverflow.com/a/17016257)

PR #399, opened 2023-06-22
Add spatial feature class

AhmetNSimsek:
Aims of this PR:
- [x] Add a feature base class `spatial` that handles features with location anchors.
- [x] Replace bounding box calculation of image features with precomputed bounding box coordinates in the configuration.
    - [x] Remove the need for ImageAnchor
- [x] Add a pointset spatial feature subclass that handles data such as big brain intensity that is delivered in coordinates and corresponding series of values.
- [ ] Refactor big brain intensity feature #377
    - [x] Return data as one dataframe of all the coordinates inside the region map
    - [x] Ensure it is deterministic, #295 
    - [x] Get the regional map in BigBrain space if the region is supported there
    - [x] Regional maps of lower-level structures should still be collected from the same space for consistency.
    - [x] Query for a higher level structure should return one result not a list of combined results for its leaves
    - [ ] Intensity calculation for the right hemisphere must be included
         - [x] Write the supporting code
         - [ ] Data source should be clarified with Konrad Wagstyl
- [x] Refactor CellDensityProfile
    - [x] Return data as one dataframe
    - [x] Allow digesting old and new preconfig
- [x] Allow extraction of a secondary base feature for selected coordinate from Pointcloud.
    - [x] Extend this with `__get_item__`
- [x] Add plotting functionality to pointcloud
    - [x] Plot values of a single coordinate
    - [x] Default plot method?


siibra/features/image/image.py
```diff
@@ -22,62 +22,32 @@
...
             description=None,  # lazy implementation below!
-            anchor=None,  # lazy implementation below!
+            anchor=anchor,  # lazy implementation below!
             datasets=datasets
         )
 
         _volume.Volume.__init__(
             self,
-            space_spec=space_spec,
+            space_spec=anchor.space.id,
```
xgui3783, line 45: I would check None'ness of anchor.space, or this line could raise

siibra/features/tabular/bigbrain_intensity_profile.py
```diff
@@ -13,13 +13,17 @@
...
+from typing import Union, List
+from ...locations import PointSet
+from .. import anchor as _anchor
+from .cortical_profile import BOUNDARIES
+import numpy as np
 
 
 class BigBrainIntensityProfile(
-    cortical_profile.CorticalProfile,
+    spatial.PointCloud,
```
xgui3783, line 26: shouldn't this also subclass CorticalProfile?

siibra/features/tabular/bigbrain_intensity_profile.py
```diff
@@ -36,28 +40,56 @@ class BigBrainIntensityProfile(
...
-            boundary_positions={
-                b: boundaries[b[0]]
-                for b in cortical_profile.CorticalProfile.BOUNDARIES
-            }
+            pointset=pointset,
+            value_headers=depths
         )
-        self.location = location
+
+    def _check_sanity(self):
```
xgui3783, line 70: ditto, see above (code repetition would mean something is fishy)

siibra/features/tabular/bigbrain_intensity_profile.py
```diff
@@ -36,28 +40,56 @@ class BigBrainIntensityProfile(
...
+        assert len(self._values) == len(self._depths)
+        assert all(0 <= d <= 1 for d in self._depths)
+        if self.boundaries_mapped:
+            assert all(0 <= d <= 1 for d in self.boundary_depths.values())
+            assert all(
+                layerpair in BOUNDARIES
+                for layerpair in self.boundary_depths.keys()
+            )
+
+    def assign_layer(self, coordinate, depth: float):
```
xgui3783, line 83: ditto, see above (code repetition would mean something is fishy)

xgui3783 (changes requested):

siibra/features/tabular/bigbrain_intensity_profile.py
AhmetNSimsek: Cortical profile will be removed completely. That is one of the things I wanted to talk about.

AhmetNSimsek: ReceptorDensityProfile uses plot, data, and _check_sanity of CortialProfile. All these will need to change for CellDensity and BigBrainIntensity under one roof (that is another thing to talk about in terms of design). Then, what is needed for ReceptorDensityProfile could just be within ReceptorDensityProfile from CortialProfile.
This is of course just one idea.

xgui3783: So your proposal is Receptor Density Profile will subclass Tabular directly?

AhmetNSimsek: Indeed

AhmetNSimsek: I see what you mean. I was thinking if ReceptorDensityProfile is the only class that is using CortialProfile elements, why not carry it there? There are a few constants Layerwise-* features use but we could just as well pull them from ReceptorDensityProfile. While BigbrainIntesityProfile (in this branch) has more or less repeated code (not the final version), it differs from CortialProfile and requires overriding the method either way. So this makes CortialProfile less useful.

I will take another look to see if I can make changes in CortialProfile for catering to a broader range and preserve the existing structure, except for adding PointCloud base to BigbrainIntensityProfile and CellDensityProfile.

xgui3783: This actually could also be breaking old code. e.g:

```python
siibra.features.get('hoc1 left', 'cortical profile')
```

would fetch receptor profile + layerwise intensity. would fail with the proposed change

AhmetNSimsek:
Code snippet to see the changes in BigBrainIntensityProfile
```python
import siibra
fts = siibra.features.get(siibra.get_region('2.9', 'occipital lobe'), siibra.features.cellular.BigBrainIntensityProfile)
print(len(fts))
f = fts[0]
print(f.data)
f.plot()
f.plot(point=67, backend='plotly')
```

AhmetNSimsek:
IMPORTANT: before merging, the commit a968ac74f388921fa2f76864057a4e57c0e26fee must be reverted. It was only made to simplify testing.

AhmetNSimsek:
This PR has been converted to draft only because it is considered to be breaking for v0.4. It might be prudent to create a new branch for v0.5 (which this branch could be the starter of) and develop this further there. This also implies the need for a v0.4 end-of-life branch.

AhmetNSimsek:
Test: time getting image section instances (after initializing maps, parcellations, and spaces).
| | No local cache | Cached on disk | 
| --- | --- | --- |
| Read bbox (a968ac74f388921fa2f76864057a4e57c0e26fee) |  (19.1s, 19.1s, 18.9s) | (0.151s, 0.178s, 0.155s)  |
| Current (4057d1a47a0a88b71bcc7433218cdb01c9a2aad9) | (19s, 18.5s, 18.7s)  | (0.189s, 0.135s, 0.085s)  |

PR #400, opened 2023-06-23
Warn of non-symmetric connectivity matrix on creation of dataframe, and allow getting row profile aside

AhmetNSimsek:


siibra/features/connectivity/regional_connectivity.py
```diff
@@ -188,24 +188,29 @@ def get_profile(
         region: Union[str, _region.Region],
         subject: str = None,
         min_connectivity: float = 0,
-        max_rows: int = None
+        max_rows: int = None,
+        direction: str = 'column'
```
xgui3783, line 192: Rather than any str, I would hint that column and row are the only valid arguments here

siibra/features/connectivity/regional_connectivity.py
```diff
@@ -188,24 +188,29 @@ def get_profile(
...
+            Regions with connectivity less than this value are discarded.
         max_rows: int, default: None
+            Max number of regions with highest connectivity.
+        direction: str, default: 'column'
+            Choose the direction of profile extraction particularly for
+            non-symmetric matrices. ('column' or 'row')
         """
         matrix = self.get_matrix(subject)
+        if direction == 'row':
+            matrix = matrix.transpose()
```
xgui3783, line 213: I would also add a check if the user supplied a non sense argument as direction, this method should raise

PR #405, opened 2023-06-23
test: add test to e2e map.publications

xgui3783:


AhmetNSimsek (approved):
LGTM
This, however, won't check for what was the core of the issue. The result was returning `[[]]` which has a length 1.

Issue #389, opened 2023-05-30
[feature request] allow connectivity data to be (more) lazily loaded


xgui3783:
as of 8d11569d15ca170ebe6c196da3fac11e7b5c37b1

code to reproduce:

```python
import siibra
fs = siibra.features.get(
  siibra.parcellations['2.9'],
  "streamline lengths"
)
f = fs[0]
f.get_matrix(f.subjects[0])
```

expected result:

only downloads the relevant csv file

actual result:

downloads the entire zip archive (~300MB)

related to https://github.com/FZJ-INM1-BDA/siibra-python/issues/190

AhmetNSimsek:
This is on my to-do list. It is a general issue how we deal with zip files. (the issue lies in `zipfile` property.)

Issue #402, opened 2023-06-23
[Refactor] ZipfileRequest and ZipfileConnector


AhmetNSimsek:
Issues:
- [ ] ZipfileConnector.search_files() cannot return duplicate file names in different folders
- [ ] ZipfileConnector downloads the whole zip file in order to pull individual files
- [ ] ZipfileRequest downloads the whole zip file in order to pull individual files

PR #406, opened 2023-06-23
Refactor ZipfileConnector and ZipfileRequest

AhmetNSimsek:
To allow more lazy loading from zipfiles. See #389 and #402

Note: Use `zipConnRefactorAdjustments` branch of siibra-configurations for testing

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 per-file-ignores = __init__.py:F401,E402
-ignore = E501,W503,F722
+ignore = E501,W503,F722, E731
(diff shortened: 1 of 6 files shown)


AhmetNSimsek:
As of 1bc82bd4e0f02ecd603ef35098ccaffb6b43b247 the requested features are achieved. However, when the average connectivity matrix is requested, it is very slow. Any ideas @xgui3783?

AhmetNSimsek:
Note: `requests.get(url, stream=True)` caches the whole zip to memory. So calling several files implies downloading the zipfile to memory several times.
Considering implementing a 3rd party zip requester. A good candidate: https://github.com/gtsystem/python-remotezip

xgui3783:
does not fix https://github.com/FZJ-INM1-BDA/siibra-python/issues/389, if it downloads the entire zip file.

AhmetNSimsek:
Stale. A new version will be created for v0.5_dev

PR #408, opened 2023-06-27
add search_datasets that returns a list of dataset metadata

AhmetNSimsek:


AhmetNSimsek:
Decided to be outside the scope of siibra-python. If needed, there is a code snippet at [brainscapes-configurations](https://jugit.fz-juelich.de/t.dickscheid/brainscapes-configurations/-/commit/45695b3b318138296fd6e2ffb187eeeda99ca10e)

PR #409, opened 2023-06-29
Get resampled template of a map instance

AhmetNSimsek:
```python
import siibra
mp = siibra.maps.COLIN27_JBA29_LABELLED
mp_img = mp.fetch()
print(mp_img.shape)

template = siibra.spaces['colin27'].get_template().fetch()
print(template.shape)

resamp_template = mp.get_resampled_template()
print(resamp_template.shape)
```
Out:
```
(256, 256, 256)
(181, 217, 181)
(256, 256, 256)
```

### DIFF
--- a/e2e/volumes/test_parcellationmap.py
+++ b/e2e/volumes/test_parcellationmap.py
@@ -9,9 +9,10 @@ from siibra.volumes.volume import Subvolume
 from itertools import product
 
 maps_to_compress = [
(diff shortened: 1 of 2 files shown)


xgui3783:
why not add the code snippit as an e2e test?

AhmetNSimsek:
> why not add the code snippit as an e2e test?

I was just doing this as a second commit 😅

PR #416, opened 2023-07-11
Migrate ebrainsFeature to EbrainsDatasetVersionV3

AhmetNSimsek:
EbrainsFeature queries have been using KG v1 connection. This PR makes general adjustments to ebrains data features around EbrainsDatasetVersionV3 queries.

### DIFF
--- a/siibra/features/dataset/ebrains.py
+++ b/siibra/features/dataset/ebrains.py
@@ -22,7 +22,11 @@ from ...retrieval.datasets import EbrainsV3DatasetVersion
 
 
 class EbrainsDataFeature(feature.Feature, category="other"):
-    def __init__(self, dataset_version_id: str, anchor: _anchor.AnatomicalAnchor):
+    def __init__(
+        self,
+        dataset_version_id: str,
+        anchor: _anchor.AnatomicalAnchor
+    ):
         feature.Feature.__init__(
             self,
             modality=None,  # lazy implementation below
--- a/siibra/features/dataset/ebrains.py
+++ b/siibra/features/dataset/ebrains.py
@@ -18,7 +18,7 @@
 from .. import anchor as _anchor
 from .. import feature
 
-from ...retrieval.datasets import EbrainsV3DatasetVersion
+from ...retrieval import datasets
 
 
 class EbrainsDataFeature(feature.Feature, category="other"):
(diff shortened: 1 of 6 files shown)


AhmetNSimsek:
TODO: Some datasets have no fullname field in their detail. This causes some features to return a None type name.
```python
import siibra
fts = siibra.features.get(siibra.get_region('wax v3', 'neocortex hippocampus'), siibra.features.other.EbrainsDataFeature)
```
@xgui3783, any ideas how to remedy this without making changes in the query?

AhmetNSimsek:
Potentially, one can add the following to provide access to the data:
```python
def get_connector(self) -> RepositoryConnector:
    return EbrainsPublicDatasetConnector(self._dataset.id)
```

siibra/livequeries/ebrains.py
```diff
@@ -114,7 +109,9 @@ def query(self, region: region.Region):
                     versioned_datasets[name][version] = dset
 
         if len(invalid_species_datasets) > 0:
-            with NamedTemporaryFile(mode='w', suffix='.txt', delete=False, encoding='utf-8') as f:
+            with NamedTemporaryFile(
+                mode="w", suffix=".txt", delete=False, encoding="utf-8"
+            ) as f:
```
xgui3783, line 114: Just a comment for now, but is it possible to turn off the automatic formatting of code, or bake it into CI?

It makes it really difficult to find which part of the code has meaningful changes, and which part is autoformatted.

AhmetNSimsek: I see what you mean. Maybe we can do a general autoformatting first? Because these are just following flake8 bases on our .flake8 config.

siibra/features/dataset/ebrains.py
```diff
@@ -18,70 +18,62 @@
 from .. import anchor as _anchor
 from .. import feature
 
-from ...retrieval import datasets
+from ...retrieval.datasets import EbrainsV3DatasetVersion
```
xgui3783, line 21: `from ...retrieval import datasets` is specifically used to prevent circular imports. I know we probably do not have circular imports here, but for future proofing, we should stick with the convention.

siibra/features/dataset/ebrains.py
```diff
@@ -18,70 +18,62 @@
...
 
 
-class EbrainsDataFeature(feature.Feature, datasets.EbrainsDataset, category='other'):
-
+class EbrainsDataFeature(feature.Feature, category="other"):
```
xgui3783, line 24: semantically I think ebrainsdatafeature should subclass `datasets.EbrainsV3DatasetVersion`

But I think maintenance wise, this makes more sense.

xgui3783 (approved):
Just the import, otherwise LGTM.

siibra/features/dataset/ebrains.py
AhmetNSimsek: I get what you mean. I was also conflicted but landed here. If the need arises, we'll bring back again. And as far as I know, this is not a breaking change.

PR #421, opened 2023-07-12
Fix the use of local config on windows

AhmetNSimsek:
Alternative to https://github.com/FZJ-INM1-BDA/siibra-python/pull/394


AhmetNSimsek:
stale

PR #425, opened 2023-08-02
Add preloading preconfigs of modules to memory

AhmetNSimsek:
EDT: Profiling results below.

### DIFF
--- a/siibra/__init__.py
+++ b/siibra/__init__.py
@@ -113,6 +113,23 @@ if "SIIBRA_CACHE_SIZE_GIB" in _os.environ:
     set_cache_size(float(_os.environ.get("SIIBRA_CACHE_SIZE_GIB")))
 
 
+def preload_configurations():
+    """
+    Preload preconfigured siibra concepts.
+
+    Siibra relies on preconfigurations that simplify integrating various
+    concepts such as parcellations, refernce spaces, and multimodal data
+    features. By preloading the instances, siibra commits all preconfigurations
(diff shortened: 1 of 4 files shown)


AhmetNSimsek:
```python
import time
import siibra
siibra.core.load_preconfigurations()
siibra.volumes.load_preconfigurations()
start = time.time()
siibra.get_map(parcellation="julich 2.9", space="MNI152")
print(time.time() - start)
```
| | WO preload | W map preload | W map and core preload |
| --- | --- | --- | --- |
| WO local cache | 44.7s  | 0.58s | 0.029s |
| W local cache | 5.6s | 0.27s | 0.027s |

xgui3783 (approved):
Looks good. 

I dont' know about the function name though.

`load_preconfigurations` -> I know in the background we quite literally loads the preconfiguration into memory. But I think it still can be called something generic, like `warmup`. This is so that in the future, we can add warm up strategies that does not involve preconfiguration

siibra/__init__.py
```diff
@@ -135,4 +152,5 @@ def __dir__():
         "vocabularies",
         "__version__",
         "cache",
+        "preload_configurations"
```
xgui3783, line 155: Wrong export

xgui3783 (approved):
Just one typo, otherwise good to go.

siibra/__init__.py
AhmetNSimsek: Thanks for catching!

PR #427, opened 2023-08-02
Maint: RepositoryConnector._decode_response

AhmetNSimsek:


siibra/retrieval/requests.py
```diff
@@ -134,7 +148,15 @@ def find_suitiable_decoder(url: str):
         else:
             return None
 
-    def _set_decoder_func(self, func):
+    def _set_decoder_func(self, func: None):
```
xgui3783, line 151:
```suggestion
    def _set_decoder_func(self, func: Callable=None):
```

siibra/retrieval/requests.py
```diff
@@ -119,7 +119,21 @@ def __init__(
...
+        url : str
+            The url or filename with extension.
+
+        Returns
+        -------
+        Callable or None
+        """
         urlpath = urllib.parse.urlsplit(url).path
         if urlpath.endswith(".gz"):
             dec = HttpRequest.find_suitiable_decoder(urlpath[:-3])
```
xgui3783, line 139: if `find_suitable_decode` returns `None` this code will break

siibra/retrieval/repositories.py
```diff
@@ -43,15 +43,8 @@ def _build_url(self, folder: str, filename: str):
         pass
 
     def _decode_response(self, response, filename):
-        # see if we find a default encoder
-        suitable_decoders = [
-            dec for sfx, dec in DECODERS.items() if filename.endswith(sfx)
```
xgui3783, line 48: is `DECODER` still used elsewhere? or the imports can also be removed?

xgui3783 (commented):
it feels to me that `find_suitiable_decoder` can graduate and be on its own, if multiple classes are requiring it.

siibra/retrieval/repositories.py
AhmetNSimsek: There are a few places but I think I can remove them as they should be resolved automatically.

siibra/retrieval/requests.py
AhmetNSimsek: Good point

Issue #350, opened 2023-04-20
[discussion] use plotly as a visualization grammar for plots to be also transferred by the api


AhmetNSimsek:

xgui3783:
[meta]

do we use discussion feature, or rely exclusive on issues?

if we do not expect to use discussion, we should disable discussion tab altogether

[/meta]

alternatives include:

- plotly

ch-schiffer:
Commenting on plotly vs. vega: I tried both `vega` (through `altair` python library) and `plotly`, and my personal favorite is `plotly`. Both offer nice interactivity, which is a huge plus over for example matplotlib, but I found `plotly` just way easier to work with and understand, and I liked the documentation better.
`altair` does its best to hide away some of the details of vega (which is a good thing in my opinion, I cannot imagine writing vega directly to be fun), but at times it still feels like writing a quite verbose specification for the plot.

AhmetNSimsek:
Decided to use `plotly`

AhmetNSimsek:
Note that `pandas` can use `plotly` as the backend visualization grammar by `pd.options.plotting.backend = "plotly"`. For simplicity and keeping requirements of siibra-python low, [a prototype implementation](https://github.com/FZJ-INM1-BDA/siibra-python/tree/implementPlotly) has been written.
If plotly is installed in the environment, the plot functions use plotly and return as a `plotly.graph_objects.Figure`. This object has `to_json()` method which can be used for `siibra-api`.
@xgui3783, when you have time, could you please test this?

xgui3783:
can you raise a draft PR already? so we can keep the discussion in the PR, rather than in this issue?

AhmetNSimsek:
As of 5004e69597d89f74bdaec7fcc184737ed71c57e7, Plotly has been implemented in general.
TODO: make sure all plots in pyplot preserve the settings of style settings of matplotlib.

AhmetNSimsek:
Continued at #429

PR #429, opened 2023-08-03
Preserve the style settings of matplotlib for pyplot

AhmetNSimsek:
#350

### DIFF
--- a/siibra/features/connectivity/regional_connectivity.py
+++ b/siibra/features/connectivity/regional_connectivity.py
@@ -32,7 +32,6 @@ except ImportError:  # support python 3.7
     from typing_extensions import Literal
 
 
-
 class RegionalConnectivity(Feature):
     """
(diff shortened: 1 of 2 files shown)


xgui3783:
For this to be adopted in siibra-explorer, the current `plot_profile` will need to be accessible via `plot` method, since siibra-api only access the plot method.

proposal: add kwarg `region`. If defined, will plot profile, rather than matrix.

AhmetNSimsek:
> For this to be adopted in siibra-explorer, the current `plot_profile` will need to be accessible via `plot` method, since siibra-api only access the plot method.
> 
> proposal: add kwarg `region`. If defined, will plot profile, rather than matrix.

I think this makes sense. However, since it is a breaking change, I propose that we leave the renaming certain plot functions to `plot` for v0.5. What do you think?

PR #430, opened 2023-08-04
Allow getting features for a list of AtlasConcepts

AhmetNSimsek:


xgui3783:
Has anyone asked for this?

xgui3783:
My concern is, since we are moving from a 1-N relationship to a M-N relationship, there are no obvious links between returned features and the supplied Atlas concepts

Issue #432, opened 2023-08-09
issue with groupby and Region


ch-schiffer:
The `assign` function returns a data frame with a column `region`, where the entries are `Region` objects. However, it is not possible to `groupby` this column, as the comparison operator `<` is not defined for `Region`  objects. This is quite inconvenient, as grouping the result by regions is common I would say.

```python
df = map.assign(some_points)
df.groupby("region") # raises an error
```

A workaround is to create a new column based on the region, e.g. `sampled_points["region_key"] = [region.key for region in sampled_points.region]`, but it is tedious to do this all the time. Maybe it is easier to just implement the comparison operator for regions, for example based on the key.

xgui3783:
Perhaps the better approach is to add comparison class methods to atlas concepts?

PR #434, opened 2023-08-14
Sort atlas concepts by their name (within the same class)

AhmetNSimsek:
Solves #432


xgui3783 (approved):
LGTM

Though I would also add some e2e/unit tests to ensure regression is covered.

ch-schiffer (approved):
Looks good to me, the script initially triggering the issue works now without  workaround.

PR #435, opened 2023-08-15
Digest unparcellated_regions from map specs

AhmetNSimsek:
Although some regions are defined in a map configuration with indices (for accuracy), they do not correspond to any regions in the corresponding region. (see https://github.com/FZJ-INM1-BDA/siibra-python/actions/runs/5740796548/job/15559446484?pr=373)
For example, on fsaverage space Julich brain 3 have indices for the following regions but they are not part of the Julich brain 3 region tree.
```json
"unparcellated regions": [
	"BrainStemMask - right hemisphere",
	"BrainStemMask - left hemisphere",
	"CorpusCallosumMask - right hemisphere",
	"CorpusCallosumMask - left hemisphere"
]
```

This PR allows digesting fields such as one above from map configurations and when a use ones to get the region object for their corresponding index, it informs the users and returns None (the same action for when there are not matches).
Also, it allows the `parcellationmap.Map.get_colormap` method to skip such regions. (While they might have a color value, the colors are currently defined in parcellation configurations.)

### DIFF
--- a/siibra/configuration/factory.py
+++ b/siibra/configuration/factory.py
@@ -281,6 +281,7 @@ class Factory:
             space_spec=spec.get("space", {}),
             parcellation_spec=spec.get("parcellation", {}),
             indices=spec.get("indices", {}),
+            unparcellated_regions=spec.get("unparcellated regions", []),
             volumes=volumes,
             shortname=spec.get("shortName", ""),
             description=spec.get("description"),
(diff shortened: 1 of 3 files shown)


AhmetNSimsek:
An alternative has been employed so that no changes in siibra-python is necessary:
> The regions that are indexed but not defined in the region tree are now listed under `unindexed_indices` in the map configurations. This field could be potentially digested by siibra-python but first the need has to be apparent.

PR #437, opened 2023-08-16
fix: parcellation.get_region exact string match

xgui3783:


xgui3783:
@AhmetNSimsek per our discussion, can you take a look?

(I have also included the expected case where the change is expected to raise in e2e test)

PR #441, opened 2023-08-25
Introduce `CompoundFeature`s to combine features such as BigBrainIntensityProfiles

AhmetNSimsek:
Follow up from https://github.com/FZJ-INM1-BDA/siibra-python/pull/399#issuecomment-1686506600, this PR allows the feature query to combine features that are allowed and compoundable while leaving the rest as is.
This PR also brings the critical changes from bigbrain livequery from #399. 

```python
import siibra
fts = siibra.features.get(siibra.get_region('julich 2.9', "occipital cortex"), siibra.features.cellular.BigBrainIntensityProfile)
for f in fts:
    print(f.name)
print(fts[0].data)
print(fts[0][5].data)
```

Also,
```python
fts = siibra.features.get(siibra.get_region('julich 2.9', "occipital cortex"), 'bigbrain')
for f in fts:
    print(f)
    print(f.name + '\n')
```
```python
<siibra.features.feature.CompoundFeature object at 0x141444d00>
CompoundFeature of 2878 BigBrainIntensityProfiles (Modified silver staining) anchored at Area hOc2 (V2, 18) left, Area hOc1 (V1, 17, CalcS) left with Set of 2878 points in the Bounding box from (-63.69,-69.55,-31.28) mm to (3.08,77.90,54.64)mm in BigBrain microscopic template (histology) space

<siibra.features.tabular.layerwise_bigbrain_intensities.LayerwiseBigBrainIntensities object at 0x1417c0a00>
LayerwiseBigBrainIntensities (Modified silver staining) anchored at Area hOc1 (V1, 17, CalcS) left

<siibra.features.tabular.layerwise_bigbrain_intensities.LayerwiseBigBrainIntensities object at 0x17aec8eb0>
LayerwiseBigBrainIntensities (Modified silver staining) anchored at Area hOc2 (V2, 18) left
```

**TODO**
- [x] `id` property of `CompoundFeature`s should reproduce the feature using `get_instance_by_id`.

EDT: updated todos based on dev meeting. Replaced output with the latest commit. 

edit 14.09.2023:
- The proposed architectural change involves creating a `CompoundFeature` class in feature.py as a subclass of `Feature` so that `CompoundFeatures` can be obtained from feature queries.
- Every feature type has a `_IS_COMPOUNDABLE`, by default `False`. Only feature types that are compoundable are combined at the end of the feature query. Every compoundable feature type must have a valid `_filter_key` property to be able to filter from the `CompoundFeature`.
- Currently, only features with the same type and modality can be compounded.
- Compounding:
    - take a list of features
    - ensure type. modality, and `filter_key` uniqeness
    - add anchors to each other to have one anchor and return the compound feature which holds the subfeatures as a property.
- `data` property and plot functionality returns the average of the subfeatures. If a `filter_key` or integer is provided, the `plot` invokes the plotting of the subfeature. This is particularly useful for connectivity matrices.
- Anchors are added by updating the list of regions and their assignment, and union of locations. Requires to have the same species. This functionality also required the implementation of the union of locations. To avoid cyclic imports and repeating code, the union is handled at the `locations` module level.
Currently (incl. #466), compoundable features are: BigBrainIntensityProfiles, CellDensityProfiles, all types of connectivity and regional activity timeseries.
(Please see https://github.com/FZJ-INM1-BDA/siibra-python/pull/466 for connectivity examples)


siibra/features/feature.py
```diff
@@ -306,7 +325,12 @@ def livequery(cls, concept: Union[region.Region, parcellation.Parcellation, spac
         return live_instances
 
     @classmethod
-    def match(cls, concept: Union[region.Region, parcellation.Parcellation, space.Space], feature_type: Union[str, Type['Feature'], list], **kwargs) -> List['Feature']:
+    def match(
+        cls,
+        concept: Union[region.Region, parcellation.Parcellation, space.Space],
+        feature_type: Union[str, Type['Feature'], list],
+        **kwargs
+    ) -> List[Union['Feature', 'CompoundFeature']]:
```
xgui3783, line 333: re: `Union['Feature', 'CompoundFeature']` since compound feature subclasses feature, this is not necessary

siibra/features/feature.py
```diff
@@ -73,14 +75,15 @@ def __init__(
         self._description = description
         self._anchor_cached = anchor
         self.datasets = datasets
+        self.is_compound = False
```
xgui3783, line 78: I wonder if this is necessary, since user could just do a `isinstance` check

siibra/locations/__init__.py
```diff
@@ -13,7 +13,71 @@
...
 # limitations under the License.
 
-from .location import WholeBrain
+from .location import WholeBrain, Location
 from .point import Point
 from .pointset import PointSet
 from .boundingbox import BoundingBox
+
+
+def override_union(loc0: 'Location', loc1: 'Location') -> 'Location':
```
xgui3783, line 22: shouldn't this be an instance method of `Location`?

siibra/features/feature.py
AhmetNSimsek: I felt the need in examples (03-007) because using `isinstance` didn't feel elegant, and thought the need might arise later as well. Should I remove it?

siibra/locations/__init__.py
AhmetNSimsek: I thought we discussed that it is better to have it at the module level (as it has access to all location types and prevents repeating code) and python allows static methods to be used as instance methods where the first argument becomes the instance.

siibra/features/feature.py
```diff
@@ -431,3 +471,165 @@ def __getattr__(self, __name: str):
...
+            self,
+            modality=modality,
+            description=f"Compound of {len(features)} {type(features[0])}",
+            anchor=combined_anchor,
+            datasets=list({ds for f in features for ds in f.datasets})
+        )
+        self._data_cached = None  # only to be used for the average
+        self._queryconcept = queryconcept
+        self.is_compound = True
+
```
xgui3783, line 516:
```suggestion
    @property
    def subfeature_index(self) -> List[Any]:
        """Indices, in addition to int, by which subfeatures can be accessed"""
        return list(self._subfeatures.keys())

```

siibra/features/feature.py
xgui3783: makes sense

siibra/features/feature.py
```diff
@@ -431,3 +471,170 @@ def __getattr__(self, __name: str):
...
+        Parameters
+        ----------
+        features
+        queryconcept
+
+        Returns
+        -------
+        List[CompoundFeature]
+        """
+        return [CompoundFeature(features, queryconcept)]
```
xgui3783, line 592:
```suggestion
        return [cls(features, queryconcept)]
```

More customary to use cls rather than actual class name. This is so, if any class subclassing compound feature in the future, and `.compound` is called, the actual cls will be used, rather than `CompoundFeature`

siibra/features/feature.py
```diff
@@ -431,3 +471,170 @@ def __getattr__(self, __name: str):
...
+
+    @classmethod
+    def get_instance_by_id(cls, feature_id: str, **kwargs):
```
xgui3783, line 595: this unfortunately does not work. it may need to be added to `Feature.get_instance_by_id` code snippet:

```python
import siibra
f = siibra.features.get(siibra.get_region("2.9", "hoc1 left"), "big brain")
f = f[1]
print(f) # prints <siibra.features.feature.CompoundFeature object at 0x7f70aff82fe0>
print(f.name) # prints CompoundFeature of 1579 BigBrainIntensityProfiles (Modified silver staining) anchored at Area hOc1 (V1, 17, CalcS) left with Set of 1579 points in the Bounding box from (-63.69,-59.94,-29.09) mm to (0.91,77.90,54.03)mm in BigBrain microscopic template (histology) space

_id = f.id
siibra.features.Feature.get_instance_by_id(_id) # raises indexerror
```

AhmetNSimsek: By 930ce1419550106d744dee883fbc95bb710fd78c this is now possible.

AhmetNSimsek:
- [ ] TODO: rebase or update from main for important fixes and the new documentation pipeline

PR #413, opened 2023-07-05
Simplify adding custom parcellation, parcellation map, and find connections with known regions

AhmetNSimsek:
Example: [examples/05_anatomical_assignment/003_custom_parcellation.py](https://github.com/FZJ-INM1-BDA/siibra-python/compare/main...feat_custom_parcellation#diff-4963f8b621cc66dc8bd6dcc2f55e3b0dd2c62a5bbd4f1ada804e294fd8014f87)

PR #442, opened 2023-08-29
Add custom parcellation maps, compute relations to preconfigured maps, find features based on relations

AhmetNSimsek:
This PR was made to be strictly additive and it is based on https://github.com/FZJ-INM1-BDA/siibra-python/pull/413.
The code to test can be found in examples/05_anatomical_assignment/003_custom_parcellation.py.

To discuss:
- If a feature type have been queried before the relation computation, the second call ignores the relations (the same issue was present in #413). Therefore, the instances are cleaned in the example before the second query. Ideally, this should not be necessary.
- How should the relations be exported for later use?
- siibra should be able to create reconfiguration files for custom maps to avoid repetition on the user side. (IMO)


AhmetNSimsek:
Marked as stale. Alternative #520

PR #444, opened 2023-08-30
Maint: ebrains datasets, private some properties and methods 

AhmetNSimsek:
This was a result of a conversation with @dickscheid on 29.08.2023. `detail` property and `parse_person()` should be private as the users have no direct use for it.

@xgui3783, would you mind writing a doc string for `_parse_person`?


xgui3783:
IMO, this would be a breaking change, since we turn public attr to private attr

Why turn them private anyway?

AhmetNSimsek:
I would normally agree but, in this case, we could not think of a reason for them to be exposed to the user. That is, it is not the same as hiding `Feature.anchor` which is frequently used. Does this impact siibra-api?

PR #445, opened 2023-08-30
General maintainance of doc examples and adding a new example for BOLD signals

AhmetNSimsek:
https://siibra-python.readthedocs.io/en/maint_doc_examples_and_add_bold/examples/03_data_features/008_functional_timeseries.html#sphx-glr-examples-03-data-features-008-functional-timeseries-py


.github/workflows/test-examples.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         # TODO
         # only run full suite on merge to main
         # run individual push on 3.7 kernel
-        python-version: ['3.7'] # [ '3.10', '3.9', '3.8', '3.7' ]
+        python-version: ['3.8'] # [ '3.10', '3.9', '3.8', '3.7' ]
```
xgui3783, line 16: I would only make this change if we officially drop support for 3.7

AhmetNSimsek: Without it, the doc does not build so there is that. It is also already in maim https://github.com/FZJ-INM1-BDA/siibra-python/commit/7f353bccaff2f10825c75b6664931c02efb35f2a
@dickscheid wants to discuss this further but for now, we don't really have much of a choice.

docs/_static/siibra.css
```diff
@@ -33,3 +33,6 @@
 	padding: 0;
 }
 
+.wy-nav-content {
+    max-width: 1000px !important;
```
xgui3783, line 37: I would use `100dvw` rather than hardcoding px value

xgui3783 (approved):
rest LGTM

.github/workflows/test-examples.yml
xgui3783: If it's already in main, why are we committing a change in this branch?

AhmetNSimsek: It was not when this branch was created and we wanted to see the example on the doc

xgui3783: ok. I will raise issue on dropping 3.7 support, but this PR is good to go IMO

AhmetNSimsek: Thank you!

PR #450, opened 2023-08-31
feat: add export to feature

### LINKED ISSUE #440: [feature request] allow features to be saved
currently, once user gets a handle on feature, it is not possible/difficult to export features outside siibra. 

proposal:

add a `export` abstract method to Feature, which all classes subclassing feature must implement. (They can choose to raise a NotImplementedError)

It could (in the order of preference):

- accept a fileIO as input, and writes a `tar.gz` or `zip` file to the file handle (similar to `json.dump`)
- accepts filename/Pathlike as an input, and writes the file directly (in this case, format can be inferred from the extension) (similar to `nibabel.save`)
- returns bytes (of zip or tar.gz), allow user(s) to determine how the file is written 
- return a list of tuple of filename, byte, allow users to choose which file to save

xgui3783:
allow feature to be exported as zip

fixes #440


AhmetNSimsek (approved):
LGTM

PR #454, opened 2023-08-31
fix: correctly use Parcellation.find_regions in anchor

AhmetNSimsek:
Adds __eq__ to species class

### DIFF
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -750,3 +750,11 @@ class Species(Enum):
 
     def __repr__(self):
         return f"{self.__class__.__name__}: {str(self)}"
+
+    def __eq__(self, other: Union['Species', str]):
+        if isinstance(other, Species):
+            return self.value == other.value
+        elif isinstance(other, str):
+            return str(self) == f"{other.lower().replace('_', ' ')}".capitalize()
+        else:
+            raise ValueError(f"Cannot compare {type(self)} with {type(other)}")
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -750,11 +750,3 @@ class Species(Enum):
 
     def __repr__(self):
         return f"{self.__class__.__name__}: {str(self)}"
-
-    def __eq__(self, other: Union['Species', str]):
-        if isinstance(other, Species):
-            return self.value == other.value
-        elif isinstance(other, str):
-            return str(self) == f"{other.lower().replace('_', ' ')}".capitalize()
-        else:
-            raise ValueError(f"Cannot compare {type(self)} with {type(other)}")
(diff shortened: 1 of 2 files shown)


siibra/commons.py
```diff
@@ -750,3 +750,11 @@ def __str__(self):
 
     def __repr__(self):
         return f"{self.__class__.__name__}: {str(self)}"
+
+    def __eq__(self, other: Union['Species', str]):
+        if isinstance(other, Species):
+            return self.value == other.value
```
xgui3783, line 756: can be just `super().__eq__(other)` I think?

siibra/features/anchor.py
```diff
@@ -188,12 +188,15 @@ def regions(self) -> Dict[Region, AssignmentQualification]:
             regions = {
                 r: AssignmentQualification.EXACT
                 for species in self.species
-                for r in Parcellation.find_regions(self._regionspec, species)
+                for r in Parcellation.find_regions(self._regionspec)
+                if r.species == species
```
xgui3783, line 192: both `r.species` and `self.species` should be of type `Species(Enum)` no?

if that's not the case, they should be remedied to be the case, rather than allowing str to Species(enum) comparison

siibra/features/anchor.py
```diff
@@ -188,12 +188,15 @@ def regions(self) -> Dict[Region, AssignmentQualification]:
...
             }
             # add more regions from possible aliases of the region spec
             for alt_species, aliases in self.region_aliases.items():
                 for regionspec, qualificationspec in aliases.items():
-                    for r in Parcellation.find_regions(regionspec, alt_species):
+                    for r in Parcellation.find_regions(regionspec):
```
xgui3783, line 197: does `Parcellation.find_regions` not take `alt_species` as a second positional argument?

AhmetNSimsek: Nope
```python
@staticmethod
def find_regions(region_spec: str, parents_only=True)
````

siibra/commons.py
AhmetNSimsek: The issue appears often enough since there are several places where species properties or attributes are stored as str instead of enum that I thought might be needed. I'll take a look and make changes accordingly

siibra/features/anchor.py
AhmetNSimsek: I agree. I'll change the initial design

siibra/commons.py
xgui3783: should we change where species is stored as `str` ? since the issue will come up again and again

AhmetNSimsek: I made a push: [anchor.region_aliases Species keys instead of str](https://github.com/FZJ-INM1-BDA/siibra-python/pull/454/commits/ca14cdf3491d6593fead141b1c9098dcf82d1136).
I thought for the rest we can fix with this logic as it pops up moving forward (use enum instead of str).

PR #455, opened 2023-08-31
Raise when invalid space spec is supplied to `get_map`. Fixes #453

### LINKED ISSUE #453: `get_map` does not raise exception on invalid `space` argument
I noticed that calling `siibra.get_map` with an invalid space argument does not raise an exception, but returns None. For example, the following line `siibra.get_map(parcellation="3.0.3", space="banana")` with an invalid space returns None instead of raising an appropriate exception. There is an error message logged to the console (which is not interpretable and hence useful in a scripted context), but no exception. I would expect the function to raise a descriptive exception (e.g., telling me that I passed the wrong space and listing possible spaces) when it receives an invalid space argument, especially since this is already the behavior of the `parcellation`  argument of the same function.

Of course I could check the return value for None, but then I would need to check for None (in case the space is wrong) __and__ catch an exception (in case the parcellation is wrong). I would suggest unifying the behavior and switching to using exceptions everywhere.

AhmetNSimsek:


siibra/core/parcellation.py
```diff
@@ -169,15 +174,22 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED, s
         if not isinstance(maptype, MapType):
             maptype = MapType[maptype.upper()]
 
+        try:
+            space_inst = parcellationmap._space[space]
```
xgui3783, line 178: shouldn't we use `siibra.spaces[space]`, rather than use the _space in `parcellationmap`?

AhmetNSimsek: I wanted to avoid importing space at parcellation.py

siibra/core/parcellation.py
```diff
@@ -169,15 +174,22 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED, s
...
+        except Exception:
```
xgui3783, line 179: also, I wouldn't use a try catch here, 

`siibra.spaces[space]` should raise a more meaningful exception, (e.g. `SpecNotFoundException`) and let user catch it themselves

siibra/volumes/parcellationmap.py
```diff
@@ -28,7 +28,7 @@
     Species,
     CompareMapsResult
 )
-from ..core import concept, space, parcellation, region as _region
+from ..core import concept, space as _space, parcellation, region as _region
```
xgui3783, line 31: why is this renaming necessary?

AhmetNSimsek: It is not part of the fix but it is a maintenance bit I wanted to do. There are several cases where `space` is used in parcellationmap.py and it will be clearer to use `_space` for the import. This stemmed from an issue I noticed on another branch (#442 ) but it is important so I wanted to make it now.

siibra/core/parcellation.py
AhmetNSimsek: Good point

siibra/volumes/parcellationmap.py
xgui3783: what was the issue where `_space` was used in favor of `space`?

If this change was used to address a different issuee, it should go into a different PR?

Or the commit message should address the issue this is trying to solve.

siibra/core/parcellation.py
```diff
@@ -169,15 +174,17 @@ def get_map(self, space=None, maptype: Union[str, MapType] = MapType.LABELLED, s
         if not isinstance(maptype, MapType):
             maptype = MapType[maptype.upper()]
 
+        space_inst = parcellationmap._space.Space.registry()[space]
```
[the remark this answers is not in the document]
AhmetNSimsek: To avoid this alltogether, I am considering reverting to original state but raise when there is no candidates instead of returning None. BUT this might break a few things; I will test.

siibra/volumes/parcellationmap.py
AhmetNSimsek: The issue arises when a variable `space` is used within `parcellationmap` in the same scope which is very likely since parcellationmaps have a property called `space`. So it is only a matter of time. I just combined it here to avoid forgetting about it. You are indeed correct by stating it should be a separate commit.

PR #460, opened 2023-09-01
Raise when get_map gets a invalid space argument

### LINKED ISSUE #453: `get_map` does not raise exception on invalid `space` argument

AhmetNSimsek:
Based on discussion on https://github.com/FZJ-INM1-BDA/siibra-python/pull/455
Fixes #453

TODO: update unit test for parcellations

### DIFF
--- a/siibra/core/parcellation.py
+++ b/siibra/core/parcellation.py
@@ -31,6 +31,9 @@ import re
 #                except Exception as e:
 #                    pass
 
+class ConflictingArgumentException(ValueError):
+    pass
+
 
 class ParcellationVersion:
     def __init__(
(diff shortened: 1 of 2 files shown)


AhmetNSimsek:
Stale

PR #470, opened 2023-09-08
Test doc v0.3

AhmetNSimsek:


AhmetNSimsek:
Closing this PR since only siibra-api v1 uses siibra-python version I using v0.3EOL. And siibra-api v1 is only partially used by siibralab which will be carried over to the next version when possible.

PR #477, opened 2023-09-19
Sort and hash bigbrain intensity profiles by coordinates

AhmetNSimsek:
Should fix https://github.com/FZJ-INM1-BDA/siibra-python/issues/295. Further testing is required.

### DIFF
--- a/siibra/features/tabular/bigbrain_intensity_profile.py
+++ b/siibra/features/tabular/bigbrain_intensity_profile.py
@@ -61,3 +61,12 @@ class BigBrainIntensityProfile(
             }
         )
         self.location = location
+
+    def __lt__(self, other):
+        return self.location.__lt__(other.location)
+
+    def __gt__(self, other):
+        return self.location.__gt__(other.location)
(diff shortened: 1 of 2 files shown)


AhmetNSimsek:
The ordering logic works for every new sesion but different sessions still shows the ordering differently. This is a problem since it is not produce reproducible ordering (the values are still the same) and `siibra-api` is stateless.

Suggested solution: employ a different ordering principle like euclidean norm as the current coordinate comparison is very limited.

xgui3783:
Imo, not just an issue for siibra API. Notebooks can't be repeated either.

AhmetNSimsek:
> Imo, not just an issue for siibra API. Notebooks can't be repeated either.

Indeed, that is what I meant by "it does not produce a reproducible ordering (the values are still the same)"

AhmetNSimsek:
A better solution found https://github.com/FZJ-INM1-BDA/siibra-python/issues/295#issuecomment-1730966308

PR #485, opened 2023-10-09
add ids for openminds species

AhmetNSimsek:


siibra/commons.py
```diff
@@ -703,18 +703,21 @@ def decode(cls, spec: Union[str, 'Species', dict], fail_if_not_successful=True):
...
+            "b8bf99e7-0914-4b65-a386-d785249725f1": 7
         }
 
         if isinstance(spec, Species):
             return spec
         elif isinstance(spec, str):
-            if spec.split('/')[-1] in MINDS_IDS:
+            if spec in MINDS_IDS:
                 return cls(MINDS_IDS[spec])
-            if spec.split('/')[-1] in OPENMINDS_IDS:
```
xgui3783, line 717: Split should be left in, in the event that we actually want to parse actual uuid from kg

AhmetNSimsek: Ah that's why it is there. I could not track the reasoning.

PR #495, opened 2023-10-23
Fix bounds check when sampling volumes with points

ch-schiffer:
#493  did not fully resolve issues with out-of-bounds points when sampling volumes (the fix only worked for integer points). I unified the code branches to handle integers, lists, and numpy arrays using the same code (less error-prone). The bounds check should now work as expected, for all valid input types.

### DIFF
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -768,26 +768,25 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
         y: Union[int, np.ndarray, List],
         z: Union[int, np.ndarray, List]
     ):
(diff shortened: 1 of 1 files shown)


xgui3783:
FYI, failed tests is due to ebrains server being down. 

they can be rerun once the server is back up. I will post an update when that happens.

AhmetNSimsek (approved):
LGTM.
(`pointindex=None` is sth desired per se but there will be changes regarding assignment results, and if only one point is supplied, it is not a problem to index the point)

xgui3783:
> FYI, failed tests is due to ebrains server being down.
> 
> they can be rerun once the server is back up. I will post an update when that happens.

server should be back up, but I see the tests are already scheduled to rerun.

Issue #489, opened 2023-10-12
siibra cache is not thread-safe


ch-schiffer:
I just encountered an error while running parallel siibra calls from many different processes (through snakemake):

```
[Errno 2] No such file or directory: '/home/cschiffer/.cache/siibra.retrieval/56458555fc977a3234ba8523b554bc2eced8d2594e2efe52462e52e23638de6b_temp.lock'
  File "/mnt/ssd1/code/bigbrain_dl_feature_analysis/snakemake/bigbrain_features/Snakefile", line 172, in __rule_cortical_layer_meshes
  File "/mnt/ssd1/code/siibra-python/siibra/__init__.py", line 26, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/core/__init__.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/core/atlas.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/core/space.py", line 20, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/locations/__init__.py", line 17, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/locations/point.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/locations/boundingbox.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/locations/pointset.py", line 18, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/__init__.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/repositories.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/requests.py", line 16, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/cache.py", line 136, in <module>
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/cache.py", line 66, in instance
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/cache.py", line 82, in run_maintenance
  File "/mnt/ssd1/code/siibra-python/siibra/retrieval/cache.py", line 82, in <listcomp>
  File "/home/cschiffer/mamba/envs/atlas/lib/python3.11/concurrent/futures/thread.py", line 58, in runRuleException:
```

[...]

PR #496, opened 2023-10-25
use threading.Lock for cache maintenance functions

AhmetNSimsek:
Solves #489

### DIFF
--- a/siibra/retrieval/cache.py
+++ b/siibra/retrieval/cache.py
@@ -18,6 +18,7 @@ import hashlib
 import os
 from appdirs import user_cache_dir
 import tempfile
+from threading import Lock
 
 from ..commons import logger, SIIBRA_CACHEDIR, SKIP_CACHEINIT_MAINTENANCE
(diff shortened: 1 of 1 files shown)


AhmetNSimsek:
TO-DO:
Use the following design scheme instead:
```python

from functools import wraps

def get_lock(filepath: str):
    def outer(fn):
        @wraps(fn)
        def inner(*args, **kwargs):
            with FileLock(filepath):
                return fn(*args, **kwargs)
        return inner
    return outer


@get_lock("access_cache")
def warm_cache():
    pass

@get_lock("access_cache")
def maintenance():
    ...
```

AhmetNSimsek:
This is not working as intended. Using https://github.com/FZJ-INM1-BDA/siibra-python/pull/496#issuecomment-1845368030 is still an option but a more comprehensive maintenance of cache.py is required. Closing this PR for an an alternative including a comprehensive maintenance and filelock.

PR #505, opened 2023-11-08
Introduce PointDistribution feature

AhmetNSimsek:


AhmetNSimsek:
Need to implement a way to transform the coordinates if the values are collected from a different resolution etc.
Use config to convey transformation data?

AhmetNSimsek:
This PR awaits a decision on the formatting of point-cloud-type data.

AhmetNSimsek:
Will be revisited for version 1.1.0 or 2.0

PR #507, opened 2023-11-09
bigbrain profiles: pull meshes and profiles from ebrains. Add dataset connection.

AhmetNSimsek:


AhmetNSimsek:
Dev meeting 14.11.23 conclusion:
Since the data has always been taken from https://github.com/kwagstyl/cortical_layers_tutorial for `BigBrainIntensityProfiles` and `LayerwiseIntensityProfiles`, the data should be fetched from there and this should be reflected on the description. This implies that if the EBRAINS dataset for these profiles is later included, they must be included with a different name.

PR #510, opened 2023-11-17
Increase intensity profile compounding speed

AhmetNSimsek:
This PR increases the speed of combining anchors of Bigbrain intensity profiles significantly since the pairwise summing of point anchors scales linearly with the number of points. Potentially, there could be 170k points if one queries for the whole big brain template. This could take up to an hour with the current methods. This PR allows the implementation of a method special to such features to quickly produce a merged anchor.

```python
import siibra
fts = siibra.features.get(siibra.get_template('bigbrain'), siibra.features.cellular.BigBrainIntensityProfile)
```

Or simply
```python
import siibra
fts = siibra.features.get(siibra.get_region('julich 2.9', 'occipital cortex'), siibra.features.cellular.BigBrainIntensityProfile)
```
speeds up from 136.9s to 0.03s (on an i5-1145G7)

### DIFF
--- a/siibra/features/tabular/bigbrain_intensity_profile.py
+++ b/siibra/features/tabular/bigbrain_intensity_profile.py
@@ -15,7 +15,9 @@
 
 from . import cortical_profile
 
-from ...locations import point
+from typing import List, TYPE_CHECKING
+if TYPE_CHECKING:
+    from ...locations import point
 
 
 class BigBrainIntensityProfile(
@@ -42,7 +44,7 @@ class BigBrainIntensityProfile(
         depths: list,
         values: list,
         boundaries: list,
-        location: point.Point
+        location: 'point.Point'
     ):
         from ..anchor import AnatomicalAnchor
         anchor = AnatomicalAnchor(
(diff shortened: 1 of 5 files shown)


siibra/livequeries/bigbrain.py
```diff
@@ -128,12 +132,16 @@ def query(self, concept: structure.BrainStructure, **kwargs) -> List[layerwise_b
...
         ]).reshape((-1, 200))
 
+        anchor = _anchor.AnatomicalAnchor(
+            location=pointset.PointSet(loader._vertices[indices, :], space='bigbrain'),
+            region=str(concept),
+            species='Homo sapiens'
+        )
         result = layerwise_bigbrain_intensities.LayerwiseBigBrainIntensities(
-            regionname=str(concept),
+            anchor=anchor,
```
AhmetNSimsek, line 141: @dickscheid I don't think the anchor of BigBrain Intensity Profiles should have a region as they are anchored to vertices only. What do you think?

siibra/features/tabular/bigbrain_intensity_profile.py
```diff
@@ -67,3 +62,19 @@ def __init__(
...
     def location(self):
         return self.anchor.location
+
+    @classmethod
+    def _merge_anchors(cls, anchors: List['AnatomicalAnchor']):
+        from ...locations.pointset import PointSet
+        from ...features.anchor import AnatomicalAnchor
+
+        location = PointSet(
+            [anchor.location for anchor in anchors],
```
xgui3783, line 72: I am assuming `PointSet` has mechanism to check `[anchor.location for anchor in anchors]` is a valid argument. Otherwise this might be dangerous.

xgui3783 (commented):
I am no longer familiar with siibra-python code base, and can only provide some rather basic comments. 

perhaps it's more appropriate to ask @dickscheid for review.

Issue #459, opened 2023-09-01
drop python 3.7 support


xgui3783:
python 3.7 EOL has just expired [1]

very soon, ci pipelines will start to fail, as the github acitons starts to cut support for 3.7

(feel free to add more to the checklist)

- [ ] remove all ci that uses 3.7
- [ ] check any optimisation that is possible with 3.8 [2]
- [ ] update setup.py on python version supported
- [ ] using `cached_property` to cache some properties

[1] https://devguide.python.org/versions/
[2] https://docs.python.org/3/whatsnew/3.8.html

hugovk:
Please see PR https://github.com/FZJ-INM1-BDA/siibra-python/pull/512.

AhmetNSimsek:
> Please see PR #512.

Thank you @hugovk. We had decided to shelf this issue for the time being, but it was forgotten to be updated. This issue will be revisited soon.

hugovk:
Hello, curious what the revisited support policy decision is?

Python 3.8 is now also EOL: https://devguide.python.org/versions/

AhmetNSimsek:
> Hello, curious what the revisited support policy decision is?
> 
> Python 3.8 is now also EOL: https://devguide.python.org/versions/

Hi @hugovk 
Version 1 of siibra-python has reached a certain maturity and we will shift our focus to version 2. This question will be revisited for version 2 depending on stakeholder needs.

PR #512, opened 2023-11-18
Drop support for EOL Python 3.7

hugovk:
Fixes https://github.com/FZJ-INM1-BDA/siibra-python/issues/459.

Also:
* Upgrade Python syntax with https://github.com/asottile/pyupgrade `--py38-plus`
* Bump GitHub Actions
* Replace deprecated logger.warn with logger.warning


xgui3783:
LGTM

Didn't actually know that in py3.8
- forward refs can be dropped
- primitive types can be used directly in typing ( no need to import typing module)
- pipe operator can be used instead of union

also, apologies. We are in an administrative slog at the moment. That is why we might be too responsive until next week or so.

hugovk:
Rebased to resolve conflicts.

hugovk:
Closing due to conflicts and the issue being shelved (https://github.com/FZJ-INM1-BDA/siibra-python/issues/459#issuecomment-1906364643). Can be re-created later when needed 👍

PR #515, opened 2023-11-20
`provider.boundingbox` -> `provider.get_boundingbox()`

AhmetNSimsek:
Solves bounding box of region's possibly resulting to be the whole space

### DIFF
--- a/siibra/volumes/volume.py
+++ b/siibra/volumes/volume.py
@@ -115,11 +115,41 @@ class Volume(location.Location):
             for srctype, prov in self._providers.items()
         }
 
-    @property
-    def boundingbox(self):
-        for provider in self._providers.values():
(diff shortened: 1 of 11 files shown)


siibra/volumes/volume.py
```diff
@@ -119,7 +119,7 @@ def concat(url: Union[str, Dict[str, str]], concat: str):
     def boundingbox(self):
```
dickscheid, line 119: shouldn't we rename this accordingly? 

If we need access to the version without clipping as a property, I suggest to introduce a property "canvas", which just calls get_bounding_box(clip=False) and returns the unclipped bounding box of all voxels in physical space. This would be a nice explicit form of the unclipped version,

dickscheid (changes requested):

PR #517, opened 2023-11-29
Maint: parc.find_regions as module method. clarify find_regions

### LINKED ISSUE #513: Placement of `Parcellation.find_regions` (and `Atlas.find_regions`) and it's functionality requires clarification
Q1: Given a parcellation instance `p`, where should `p.find_regions` search regions for?
Q2: If the functionality is preserved, wouldn't it be more suitable if it is a method of `parcellation.py`? (Goal is to prevent access to the method from an instance since `find` is the designated search method.)

AhmetNSimsek:
This PR addresses the following issues/points ([see](https://github.com/FZJ-INM1-BDA/siibra-python/issues/513)):
- `parcellation.find_regions()` was a static method of `Parcellation` but it searches through all parcellations in the registry. For the user, when they have a parcellation instance `parc`, there is no clear difference between `parc.find_regions()` and `parc.find()` while they should be using `find` to search within the parcellation. (`find` is a method of `Region`).
- There is also `atlas.find_regions()` as an instance method. It behaves differently than  `parcellation.find_regions()`. Needs to be clarified.
- Finally, instead of using caching with a dictionary with keys of kwarg tuples, why not use `cache` property?
    - `lru_cache` might be necessary for long-running systems like siibra-api

(Please see the discussions on the code changes for further details)

Fixes #513


siibra/core/parcellation.py
```diff
@@ -364,3 +326,45 @@ def __lt__(self, other):
...
+        itself might not match the specification.
+
+    Returns
+    -------
+    list[Region]
+        list of regions matching to the regionspec
+    """
+    key = (regionspec, filter_children, find_topmost)
+    if regionspec not in _FIND_REGIONS_CACHE:
+        _FIND_REGIONS_CACHE[key] = []
```
xgui3783, line 361: I would use `defaultdict` rather than explicitly assign value

siibra/core/parcellation.py
```diff
@@ -364,3 +326,45 @@ def __lt__(self, other):
...
+        for p in Parcellation.registry():
+            _FIND_REGIONS_CACHE[key].extend(
```
xgui3783, line 363: is this process so slow that we need to cache the result?

I have concerns on the RAM implication on long running processes (*cough e.g. server * cough) that we willy nilly cache everything. 

I wonder if we should start to use an LRU or something

xgui3783 (approved):
Mostly fine. 

I have questions where find_region should still be in `parcellation.py`

and the aggressive caching strategy we are using everywhere.

siibra/core/parcellation.py
```diff
@@ -364,3 +326,45 @@ def __lt__(self, other):
             )
             return self.name < other.name
         return self.version.__lt__(other.version)
+
+
+def find_regions(
```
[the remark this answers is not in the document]
AhmetNSimsek: Rather than internal use, it is also exposed as `siibra.find_regions()`. We can potentially use `region.find()` in anchor. I think the original implementation also had species as a keyword argument, which made more sense to use there.

It needs access to parcellation.registry so I think it is better in parcellation module.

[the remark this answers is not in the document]
xgui3783: The fact that it needs `Parcellation.registry` is not relevant to users. 

If I am a user, and expect to find a method called `find_regions` (which, in fact, does not depend on any parcellations), I would expect it to find it under `region.py` module. 

I do like the idea of a class/static method under `Region` class, which is proxy to this method (Indeed, I think rather than it being a standalone method, I think it makes more sense to be a static/class method of `Region` class)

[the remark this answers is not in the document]
AhmetNSimsek: Hmm then it would still be accessible from region and parcellation instances. We can carry it to the region module but if we are going through all parcellations, why shouldn't it be under the parcellation module? And this also creates less confusion given that `Region` class has `find` method.

Since there is a also a similar function in `atlas`, which goes through all parcellations in an atlas, I think the design is reasonable. But we might want to discuss this design with @dickscheid , who initially implemented it.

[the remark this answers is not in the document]
xgui3783: > We can carry it to the region module but if we are going through all parcellations, why shouldn't it be under the parcellation module?

But the user doesn't know that. (nor should they)

The promise of `find_region`, at least to me, is: you provide a string, and I will find you all regions which fits this string.

Whether, internally, we go through all parcellations, and iteratively, go through all regions for that parcellation, OR, we some how have a region instance table, that is not relevant to the user. 

Which is why I say, either keep it private (prefix with `_`) or it should go in regions module.

siibra/core/parcellation.py
xgui3783: looks good

siibra/core/parcellation.py
```diff
@@ -364,3 +326,44 @@ def __lt__(self, other):
             )
             return self.name < other.name
         return self.version.__lt__(other.version)
+
+
+@lru_cache(maxsize=128)
```
AhmetNSimsek, line 331: TD: investigate how `lru_cache` hashes the inputs to compare with dictionary caching

AhmetNSimsek: https://docs.python.org/3/library/functools.html#functools.lru_cache
"Since a dictionary is used to cache results, the positional and keyword arguments to the function must be [hashable](https://docs.python.org/3/glossary.html#term-hashable)."

Since the kwargs of `find_regions are either a string a boolean, they are washable. So essentially, the key of the old implementation (`key = (regionspec, filter_children, find_topmost)`) is used but hashed.

dickscheid:
Just to explain the original logic - there where different entry points to have different scopes:
- search any region known to siibra was thought as a class method, searching through all instances of parcellations at runtime (maybe it was implemented as static which I agree was not proper then)
- search only the regions linked to a species was implemented as an instance method of atlas
- search regions in one particular parcellation was implemented as an instance method of the parcellation (implemented as  `find` in Region, since region is the parent class which holds the subtree)

I think the logic is clear, but I understand that the automatic exposure of the class method `Parcellation.find_regions` for Parcellation instances leads to confusion.

siibra/core/parcellation.py
```diff
@@ -364,3 +326,44 @@ def __lt__(self, other):
...
+def find_regions(
+    regionspec: str,
+    filter_children=True,
+    find_topmost=False
+):
+    """
+    Find regions that match the given region specification in the subtree
```
dickscheid, line 338: This comment doesn't make sense, what is the subtree here at the module level? 

should just be "find regions across all parcellation instances"?

siibra/core/parcellation.py
```diff
@@ -364,3 +326,44 @@ def __lt__(self, other):
...
+    headed by each parcellation in the registry.
+
+    Parameters
+    ----------
+    regionspec: str, regex, Region
+        - a string with a possibly inexact name (matched both against the name and the identifier key)
+        - a string in '/pattern/flags' format to use regex search (acceptable flags: aiLmsux)
```
dickscheid, line 345: can we add a link to some documentation about the flags here?

AhmetNSimsek:
> Just to explain the original logic - there were different entry points for different scopes:
> 
>     * search any region known to siibra was dessigned as a class method `Parcellation.find_regions()`, searching through all instances of parcellations at runtime (maybe it was implemented as static which I agree was not proper)
> 
>     * search only the regions linked to a species was implemented as an instance method of Atlas, which holds the parcellations of that species
> 
>     * search regions in one particular parcellation only was implemented as an instance method of the parcellation, actually implemented as  `find` in Region, since region is the parent class which holds the subtree and allows for a recursive implementation.
> 
> 
> I think the logic is clear, but I understand that the automatic exposure of the class method `Parcellation.find_regions` for Parcellation instances leads to confusion.

Thank you. Then, IMO, we should make `find_regions` a module level method in parcellation.py or if it is static or class method, it should be hidden and only be forwarded with `siibra.find_regions`. I'd prefer the first as I do not see any reason for such a method to be class or instance method.

siibra/core/region.py
```diff
@@ -179,6 +179,8 @@ def parcellation(self):
     def species(self):
         # lazy request of the root parcellation's species
         if self._species_cached is None:
+            if self.parcellation is None:
+                return None
```
dickscheid, line 183: this looks like an inconsistent state to me. Shouldn't we always have a species defined, and do we accept regions not linked to a parcellation?  I would rather raise an Exception if that happens.

siibra/core/parcellation.py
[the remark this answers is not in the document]
dickscheid: I agree that this is better placed here at the module level to avoid the confusion with Region.find()

siibra/core/parcellation.py
AhmetNSimsek: I thought " headed by each parcellation in the registry." was clear enough but we can fix it after we agree on the usage and function

siibra/core/region.py
AhmetNSimsek: TBH I do not remember why I implemented this. I will check

siibra/core/region.py
```diff
@@ -264,6 +266,7 @@ def includes(self, region):
         """
         return region == self or region in self.descendants
 
+    @lru_cache(maxsize=3)
```
xgui3783, line 269: `lru_cache` on member function will, retain the last three instances Region that called the `find` instance method, even if they go out of scope.

one can validate this by:

```python
from functools import lru_cache
from gc import collect

class B:
    def __init__(self, x: str) -> None:
        self.x = x
    
    @lru_cache(maxsize=3)
    def foo(self):
        print("calling foo")
        return self.x
    
    def __del__(self):
        print("del", self.x)

b = B("0")
b1 = B("1")
b2 = B("2")
b3 = B("3")
print(b.foo(), b1.foo(), b2.foo(), b3.foo())
print(b1.foo(), b2.foo(), b3.foo(), b.foo(), )
b1 = b2 = b3 = b = 1

```

via interactive mode. (only one instance will be deleted)

three most recent instances is not a huge issue, but it should be monitored.

The ideal work around is, IMO:

```python
from functools import lru_cache
from gc import collect

class B:
    def __init__(self, x: str) -> None:
        self.x = x
        self.foo = lru_cache(maxsize=3)(self.foo)
    
    def foo(self):
        print("calling foo")
        return self.x
    
    def __del__(self):
        print("del", self.x)

b = B("0")
b1 = B("1")
b2 = B("2")
b3 = B("3")
print(b.foo(), b1.foo(), b2.foo(), b3.foo())
print(b1.foo(), b2.foo(), b3.foo(), b.foo(), )
b1 = b2 = b3 = b = 1


```

PR #520, opened 2023-11-30
Assign maps and compute relations between parcellations

AhmetNSimsek:


xgui3783:
hmm, call me skeptical, but 

1/ is map assign so slow that we needed to optimize the performance and 
2/ does it actually increase performance?, can we do some `time` before and after?

AhmetNSimsek:
> hmm, call me skeptical, but
> 
> 1/ is map assign so slow that we needed to optimize the performance and 2/ does it actually increase performance?, can we do some `time` before and after?

I went through the same questions :) This was a very quick implementation and only tested by testing the code. I haven't tested with `time` yet because I would like to polish some other aspects regarding map assignments and relation computation at the moment, and devoting time on these.

xgui3783:
> > hmm, call me skeptical, but
> > 1/ is map assign so slow that we needed to optimize the performance and 2/ does it actually increase performance?, can we do some `time` before and after?
> 
> I went through the same questions :) This was a very quick implementation and only tested by testing the code. I haven't tested with `time` yet because I would like to polish some other aspects regarding map assignments and relation computation at the moment, and devoting time on these.

Just remember, the more code we write the more code we have to maintain <- also why, unless there is a compelling metrics to justify optimization, and/or there are specific request, I would be skeptical.

AhmetNSimsek:
Since this is sth that can be achieved by a short code, it is deemed unnecessary.

PR #525, opened 2023-12-15
Maint sparseindex

AhmetNSimsek:
TODO:
- [x] create a data proxy folder for version 1 specifically
- [x] regions should be fetchable without forming sparseindex. use sparse index for assignments only


siibra/retrieval/requests.py
```diff
@@ -65,6 +65,9 @@
     ".zip": lambda b: ZipFile(BytesIO(b)),
     ".png": lambda b: skimage_io.imread(BytesIO(b)),
     ".npy": lambda b: np.load(BytesIO(b)),
+    "sparseindex.probs.txt": lambda b: b.decode('utf-8').strip().splitlines(),
+    "sparseindex.bboxes.txt": lambda b: b.decode('utf-8').strip().splitlines(),
+    "sparseindex.voxels.nii": lambda b: Nifti1Image.from_bytes(b),
```
xgui3783, line 70: probably not needed, since it's handled by ".nii" case

siibra/retrieval/requests.py
```diff
@@ -94,9 +97,13 @@ def find_suitiable_decoder(url: str) -> Callable:
...
     ]
-    if len(suitable_decoders) > 0:
-        assert len(suitable_decoders) == 1
+    if len(suitable_decoders) == 1:
         return suitable_decoders[0]
+    elif len(suitable_decoders) > 1:
+        most_sutiable_suffix = max(
+            [sfx for sfx in DECODERS.keys() if urlpath.endswith(sfx)]
+        )
+        return DECODERS[most_sutiable_suffix]
```
xgui3783, line 106: I am not a huge fan of this approach. 

(aside: I am not a huge fan of trying to infer decoder via filename to begin with, especially since the content server send may or maynot have anything to do with the filename/extension etc)

This introduce the possibility (no matter how small) of collision (i.e. a `.txt` file where user/we accidentally name it `foobar.sparseindex.probs.txt`, but has nothing to with sparseindex we are trying to introduce here. Expecting it to be parsed by the `.txt` parser (which, as an aside, tries to parse as space delimited csv, I would not have guess for a second)

I would argue either create a new custom extension (call it `.txt.probs.sparseindex` ?) to avoid all possibility of misunderstanding/misparsing.

xgui3783 (commented):
I have an issue with the name of the files and how decoders are found. Others looks solid.

siibra/volumes/sparsemap.py
```diff
@@ -111,85 +115,121 @@ def mapped_voxels(self, volume: int):
...
         Parameters
         ----------
-        prefix: str
-            Prefix of the filenames.
+        filepath_or_url: str
+            Path/url to the SparseIndex files
+            (eg. https://url_to_files/basefilename):
+            - basefilename.sparseindex_v1.probs.txt.gz
+            - basefilename.sparseindex_v1.bboxes.txt.gz
+            - basefilename.sparseindex_v1.voxels.nii.gz
```
dickscheid, line 130: the version and type of the text file could be better reflected in the first line of the file than in the name. But I think we will reconsider this anyway in siibra-python v2, so I'd go with minimal efforts so far.

For siibra-python v2, creation of sparse indices should be a reproducible script, leading to a proper data release, ultimately with automatic curation and a good data descriptor.

dickscheid (approved):
I didn't check all code but it looks reasonable to me. once merged we need to test thoroughly that probabilistic assignment behaves identical to previous versions, especially using the tutorial notebooks that do assignment.

Issue #526, opened 2023-12-20
Assignment of uncertain points to labelled maps not working


dickscheid:
Reported by a user:

When assigning an image or an uncertain point (which siibra then models as an image of a 3D Gaussian) to a labelled map such as the Julich-brain maximum probability map, the resulting assignment table is empty. Assignment to statistical maps works. (applies to pre-1.0 releases, did not verify in 1.0 yet)

AhmetNSimsek:
Could you please share the code, or location along with the julich brain version used?

PR #531, opened 2024-01-11
Fix assigning uncertain points to labelled maps #526

AhmetNSimsek:
During refactoring (https://github.com/FZJ-INM1-BDA/siibra-python/pull/480/commits/6fe6fa6f4955342cdd0a76a2076d3594493442b9), image resampling was forgotten to be replaced. This fixes volume assignments to labelled maps.

### DIFF
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -817,10 +817,8 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
         if isinstance(item, pointset.PointSet):
             return self._assign_points(item, lower_threshold)
         if isinstance(item, _volume.Volume):
-            image = item.fetch()
             return self._assign_volume(
-                imgdata=np.asanyarray(image.dataobj),
-                imgaffine=image.affine,
+                queryvolume=item,
                 lower_threshold=lower_threshold,
                 minsize_voxel=minsize_voxel,
                 **kwargs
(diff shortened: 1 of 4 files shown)


siibra/volumes/parcellationmap.py
```diff
@@ -1057,22 +1055,25 @@ def _assign_points(self, points: pointset.PointSet, lower_threshold: float) -> L
                 shift[:3, -1] = xyz_vox[:3, 0] - r
                 # build niftiimage with the Gaussian blob,
                 # then recurse into this method with the image input
-                W = _volume.from_array(
+                gaus_kernel = _volume.from_array(
```
dickscheid, line 1058: just a name, but should be "gaussian_kernel"

siibra/volumes/parcellationmap.py
```diff
@@ -1108,27 +1113,33 @@ def _assign_volume(
...
-                    else:
-                        targetdata = (vol_data == index.label).astype('uint8')
+                region_map = self.fetch(index=index)
+                region_map_arr = np.asanyarray(region_map.dataobj)
+                # TODO: why resmapling was removed during refactoring? Needs to be addressed.
+                # imgdata_res = resample_array_to_array(
+                #     np.asanyarray(queryimg.dataobj),
+                #     queryimg.affine,
+                #     region_map_arr,
+                #     region_map.affine
```
dickscheid, line 1128: I suggest to catch different voxel coordinate systems here, ie. require same shape, space and affine and otherwise throw an exception in order not to run into this issue silently.
I agree we should resample, but I think it was not yet properly implemented anyways for all cases.

PR #537, opened 2024-01-17
refactor: warm cache

xgui3783:


siibra/retrieval/cache.py
```diff
@@ -137,3 +141,76 @@ def build_filename(self, str_rep: str, suffix=None):
...
 
 CACHE = Cache.instance()
+
+
+class WarmupLevel(int, Enum):
+    INSTANCE = 1
+    DATA = 5
+
+
+class WarmupRegException(Exception): pass
```
AhmetNSimsek, line 151: Could you carry this to `siibra.exceptions`?

siibra/retrieval/cache.py
```diff
@@ -137,3 +141,76 @@ def build_filename(self, str_rep: str, suffix=None):
...
+
+
+class WarmupParam(NamedTuple):
```
AhmetNSimsek, line 154: Could this class exist under `Warmup`?

siibra/retrieval/cache.py
```diff
@@ -137,3 +141,76 @@ def build_filename(self, str_rep: str, suffix=None):
...
+
+    @staticmethod
+    def fn_eql(wrapped_fn, original_fn):
+        return wrapped_fn is original_fn or wrapped_fn.__wrapped__ is original_fn
+
+    @classmethod
+    def is_registered(cls, fn):
+        return len([warmup_fn.fn
+                    for warmup_fn in cls._warmup_fns
+                    if cls.fn_eql(warmup_fn.fn, fn) ]) > 0
```
AhmetNSimsek, line 172: I believe this is not Flake8 compatabile
```python
def is_registered(cls, fn):
    return len([
        warmup_fn.fn
        for warmup_fn in cls._warmup_fns
        if cls.fn_eql(warmup_fn.fn, fn)
    ]) > 0
```

siibra/volumes/parcellationmap.py
```diff
@@ -1154,6 +1154,8 @@ def _assign_volume(
         return assignments
 
 
+cache.Warmup.register_warmup_fn()(Map.registry)
```
AhmetNSimsek, line 1157: `Map` is an `AtlasConcept`. I am not sure this is necessary

AhmetNSimsek (commented):
A few notes but otherwise LGTM

siibra/retrieval/cache.py
xgui3783: messes up the referencing.

I can make this private?

AhmetNSimsek: Let us leave like this for now then. @dickscheid has a design request on such small classes but we don't have a solution yet.

siibra/retrieval/cache.py
xgui3783: will fix flake and properly add it to the test

xgui3783: perhaps both are PEP valid? since it's flake passed.

xgui3783:
@AhmetNSimsek can you take a look again when you get a chance?

I fixed the tests, and properly added flake

PR #543, opened 2024-01-22
Use Github repo connector for configurations

AhmetNSimsek:
Solves #542

### DIFF
--- a/siibra/retrieval/repositories.py
+++ b/siibra/retrieval/repositories.py
@@ -43,7 +43,7 @@ class RepositoryConnector(ABC):
         self.base_url = base_url
 
     @abstractmethod
-    def search_files(folder: str, suffix: str, recursive: bool = False) -> List[str]:
+    def search_files(self, folder: str, suffix: str, recursive: bool = False) -> List[str]:
         pass
 
     @abstractmethod
@@ -127,6 +127,10 @@ class LocalFileRepository(RepositoryConnector):
             with open(self.url, 'rb') as f:
                 return self.func(f.read())
 
+    def get(self, filename, folder="", decode_func=None):
+        """Get a file right away."""
+        return self.get_loader(filename, folder, decode_func).data
+
     def get_loader(self, filename, folder="", decode_func=None):
         """Get a lazy loader for a file, for loading data
         only once loader.data is accessed."""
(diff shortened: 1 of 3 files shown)


AhmetNSimsek:
@xgui3783, I am also considering refactoring `GitlabConnector`. Specifically, how we handle archive mode.

siibra/retrieval/repositories.py
```diff
@@ -154,6 +174,104 @@ def __eq__(self, other):
...
+        ------
+        RuntimeError
+            If branch test could not find the reftag in the repo's list of tags
+            and branches.
+        """
+        RepositoryConnector.__init__(
+            self,
+            base_url=f"https://api.github.com/repos/{owner}/{repo}"
+        )
+        assert reftag
```
xgui3783, line 210: more meaning ful assert error message

siibra/retrieval/repositories.py
```diff
@@ -154,6 +174,104 @@ def __eq__(self, other):
...
+
+    def get_loader(self, filename, folder="", decode_func=None):
+        if not self.archive_mode:
+            return super().get_loader(filename, folder, decode_func)
+        else:
+            self._archive()
+            return self._archive_conn.get_loader(filename, folder, decode_func)
+
+    def _archive(self):
+        if self._archive_conn is None:
```
xgui3783, line 258: better way IMO is to check if `archive_directory` exists

So that when user quits session and rejoins session, they do not need to download the archive again.

siibra/retrieval/repositories.py
```diff
@@ -154,6 +174,104 @@ def __eq__(self, other):
...
+            archive_directory = CACHE.build_filename(self.base_url + self.reftag)
+            if not os.path.isdir(archive_directory):
+                import tarfile
+                from io import BytesIO
+
+                resp = requests.get(f"{self.base_url}/tarball/{self.reftag}", stream=True)
+                resp.raise_for_status()
+
```
xgui3783, line 266: probably easier to just use `siibra.retrieval.HTTPRequest`, no? since it also keeps a cache

siibra/retrieval/repositories.py
```diff
@@ -127,6 +143,10 @@ def data(self):
             with open(self.url, 'rb') as f:
                 return self.func(f.read())
 
+    def get(self, filename, folder="", decode_func=None):
+        """Get a file right away."""
+        return self.get_loader(filename, folder, decode_func).data
+
```
xgui3783, line 149: doesn't the parent class (`RepositoryConnector`) already implement this?)

xgui3783 (commented):
Looks good. just a few comments

siibra/retrieval/repositories.py
AhmetNSimsek: Hmm good point. I was mimicking the implementation in `GitlabConnector`.
Then, I will also change that (if you don't disagree).

siibra/retrieval/repositories.py
AhmetNSimsek: You are right. I had a bug that it tried to create an `HttpRequest` instead of a `FileLoader`. But there might be another reason; I'll investigate.

PR #544, opened 2024-01-22
fix: info uses siibra.retrieval module for local file cache

### LINKED ISSUE #530: [bug] features.get should not allow fuzzy matches, unless specified
as of `siibra==1.0a4`, feature query via bounding box allow space warping, without explicitly setting any flags. This, IMO, is the incorrect behavior. 

sample code:

```python
import siibra
from siibra.locations.boundingbox import BoundingBox

pts = [[-19.99191860496405,-17.981918604964047,-1.6956886049640476],[19.99191860496405,22.001918604964047,38.28814860496405]]
bbox = BoundingBox(*pts, "mni152")

assert len(siibra.features.get(bbox, "CellbodyStainedSection")) == 0 # fails, 33 != 0
```

this has the comedic aftermath of allow big brain images to be queried in mni space (and so on)

Screenshot 2024-01-10 164756

xgui3783:
fixes #530 , specifically https://github.com/FZJ-INM1-BDA/siibra-python/issues/530#issuecomment-1904333095

### DIFF
--- a/siibra/volumes/providers/neuroglancer.py
+++ b/siibra/volumes/providers/neuroglancer.py
@@ -235,6 +235,7 @@ class NeuroglancerVolume:
     def transform_nm(self, val):
         self._transform_nm = val
 
+    @property
     def io(self):
         if self._io is None:
             accessor = HttpAccessor(self.url)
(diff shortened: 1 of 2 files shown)


siibra/volumes/providers/neuroglancer.py
```diff
@@ -233,35 +235,39 @@ def transform_nm(self):
     def transform_nm(self, val):
         self._transform_nm = val
 
+    @cached_property
```
AhmetNSimsek, line 238: I think this is another reason to drop python 3.7 but for now could you cache with `self._io`?

AhmetNSimsek (approved):

AhmetNSimsek:
The tests that are failing are due to Allen Institue being down

PR #553, opened 2024-02-01
Experimental: feature refactor using attributes

xgui3783:
[design document](https://gifted-closet-6a9.notion.site/siibra-python-feature-redesign-ed20247363c34a8c936e5b63960dc140)

in discussion with @dickscheid and @AhmetNSimsek on the refactoring of features. 

(from previous meeting notes)

- Features are class based, which means new feature likely result in more engineering work (adding new class. See Tracer connectivity, streamline connectivity etc)

Pseudo code

```python
import siibra
from abc import ABC, abstractproperty
from collections import defaultdict

cursor: "QueryCursor" = siibra.feature.get("foo", "bar")


class Intent(ABC):
    type: str="intent"
    
    @abstractproperty
    def key(self):
        raise NotImplementedError
    
    @abstractproperty
    def value(self):
        raise NotImplementedError
    
    def filter(self, *args, **kwargs) -> bool:
        """Filter method for intent. Subclasses can overwrite 
        on how the parent feature should filter the feature."""
        return True


class Feature:
    id: str
    name: str
    desc: str
    intents: list[Intent]

    def filter(self, *args, **kwargs) -> bool:
        """Proxy method checks all intents. Shortcircuits
				if any returns false. """
        for intent in self.intents:
            if not intent.filter(*args, **kwargs):
                return False
        return True


class QueryCursor:
    
    def __init__(self, subfeatures: list[Feature]):
        self.subfeatures = subfeatures
        self.cursor = 0
        self._cached_attributes = None

    def seek(self, idx: int):
        self.cursor = idx
    
    def next(self):
        if self.cursor >= len(self):
            raise StopIteration
        f = self.subfeatures[self.cursor]
        self.cursor += 1
        return f
    
    def __iter__(self):
        yield self.next()
    
    def __len__(self):
        return len(self.subfeatures)
    
    def __getitem__(self, key: slice | int):
        return self.subfeatures[key]
    
    @property
    def attributes(self) -> dict[str, set[str]]:
        if self._cached_attributes is None:
            return_val = defaultdict(set)
            for f in self.subfeatures:
                for intent in f.intents:
                    return_val[intent.key].add(intent.value)
            self._cached_attributes = return_val
        return self._cached_attributes

    def filter(self, *args, **kwargs) -> "QueryCursor":
        filtered_features = [f for f in self.subfeatures if f.filter(*args, **kwargs)]
        return QueryCursor(filtered_features)
```

coincidentally similar to [entity component system](https://en.m.wikipedia.org/wiki/Entity_component_system)

to consider:

- do we allow Component Inheritence?
- do we allow Component member methods?


siibra/features_beta/attributes/meta_attributes.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@dataclass
+class ModalityAttribute(MetaAttribute, schema="siibra/attr/meta/modality"):
+    name: str
+
+    def matches(self, *args, modality=None, **kwargs):
+        if isinstance(modality, str) and all_words_in_string(modality, self.name):
+            return True
+        return super().matches(*args, modality=modality, **kwargs)
```
xgui3783, line 27: In theory, I have no issue with exposing a `def match(self, *args, **kwargs): bool` method to user/developer, I believe the implementation should be calling to a global/static method, (same can be said for `.data` computed attribute. 

mainly, caching instance bound method/property has lots of caveats , e.g. https://joblib.readthedocs.io/en/latest/memory.html#gotchas

siibra/features_beta/feature.py
```diff
@@ -0,0 +1,76 @@
...
+    """ A multimodal data feature characterized by a set of attributes.
+    """
+    schema = "siibra/feature/v0.2"
+    id: str = field(default_factory=lambda: str(uuid.uuid4()))
+    attributes: List["attributes.Attribute"] = field(default_factory=list)
+    name: str = field(default=None)
+
+    def __post_init__(self):
+        # Construct nested FeatureAttribute objects from their specs
+        for i, att in enumerate(self.attributes):
```
xgui3783, line 29: We don't need the index, do we? We can just iterate over `self.attributes` ?

xgui3783:
@dickscheid I have again added new commit, which, paired with configuration commit of 24d547c068237c4d3c808ea5f62a290797e03622 should work with VOI. 

related to the over zoom discussion, I have elected to not remove `VolumeProvider`, but instead, use existing `VolumeProvider` to simplify how `VoiDataAttribute` fetches volume.

My reasoning is that volume provider is also used by map, and template. Rewriting it would have a lot of downstream effects. For this rewrite, we should keep the scope limited to features. 

We can certain add a TODO to revisit on how volume is fetched, once we finish this rewrite.

siibra/features_beta/attributes/base.py
```diff
@@ -0,0 +1,42 @@
+from dataclasses import dataclass, field
+from typing import List, Any
+from ...commons import logger
+
+SCHEMAS = {}
+
+
+@dataclass(kw_only=True)
```
AhmetNSimsek, line 8: kw_only was introduced with 3.10. We should think if an alternative.

xgui3783:
todo for 2024 05 06

- [ ] merge main -> this branch @AhmetNSimsek 
- [ ] fix `kw_only` @AhmetNSimsek

xgui3783:
superceded by https://github.com/FZJ-INM1-BDA/siibra-python/pull/594

PR #554, opened 2024-02-02
Feat map to zip

AhmetNSimsek:
Save volumes and parcellation maps to a zip file. See e2e/volumes/test_tozip.py for example code.

Notes:
- Saves meshes as .gii
- Each volume of a map returns a readme.
- It was decided in [dev meeting on 2024.01.25](https://github.com/FZJ-INM1-BDA/siibra-development/blob/main/meetings/2024/20240125.md) that there should be a level before calling `fetch`, i.e., `get_volume` so that the over arching design pattern remains (`fetch` takes out user to inter-operable data types). However, this created a lot of overhead which is not expected when getting a volume object should not cause. It also made the code more complicated. (see https://github.com/FZJ-INM1-BDA/siibra-python/pull/554/commits/6867cff79fea8572ae6b37b7590c17515cc33616)

Unit test test/retrieval/test_retrieval_download_file.py is failing due to data-proxy being down.

### DIFF
--- a/e2e/volumes/test_tozip.py
+++ b/e2e/volumes/test_tozip.py
@@ -2,7 +2,7 @@ import pytest
 import siibra
 
 from siibra import MapType
-from siibra.volumes import Map
+from siibra.volumes import Map, Volume
 
 from zipfile import ZipFile
 import os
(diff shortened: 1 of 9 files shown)


dickscheid:
@AhmetNSimsek I need more guidance to respond to this issue.

- if we wanted it once, did the situation change in any way so that we ask ourselves whether or not it would still be done?
- what are advantages and disadvantages of doing it?
- is it a matter of merging this PR or are there additional efforts implied to finish it?

AhmetNSimsek:
> In general I do think we should implement download of zips, especially since it is already available in siibra-explorer and I think the functionality should be provided from siibra-python (not be implemented twice). I do not consider it urgent. If the explorer-functionality is so far independent, I am happy to reconsider it for siibra-python v2.
> 
> @AhmetNSimsek I need more guidance on this issue.
> 
>     * In how far did the situation change so that we need to reconsider our past agreement to implement it?

When the explorer was switching to version 1 of siibra-python, this was a must. But now that the explorer will switch from version 0.4 to version 2 directly, it was not a big necessity.

>     * what are advantages and disadvantages?

This PR allows downloading maps from siibra-python, which would be a good addition, but not sure it is a necessity. siibra-python version 2 handles and covers several points (see PR description) this PR solves better. 

>     * is it a matter of merging this PR or are there additional efforts implied to finish it?

Currently, this draft utilizes `Subvolume` class to handle getting regional masks. This needs to be updated to use another class (`ExtractedVolume`?). Then it can be merged.

AhmetNSimsek:
Stale. https://github.com/FZJ-INM1-BDA/siibra-python/pull/624 solves most of the issues better but does not include `to_zip` method. Since it is rather modular, it can be taken from this PR in the future.

Issue #551, opened 2024-01-26
Neuroglancer fragments cannot be merged


AhmetNSimsek:
This causes maps such as cortical layers on Bigbrain to fail when fetched without a fragment being provided.

AhmetNSimsek:
solved as of 2d7a0e49ca793e1d41ea3833f78fc2d94935eb78

PR #561, opened 2024-02-08
Introduce merging Neuroglancer fragments

AhmetNSimsek:
When a map with fragmented Neuroglancer volumes is requested without specifying a fragment, the fetch fails as merging of fragments for Neuroglancer volumes was not implemented (#551). Test with:

```python
from nilearn import plotting
import siibra
mp = siibra.get_map('cortical layers', space='bigbrain')
res = 0.64
img = mp.fetch(resolution_mm=res)
bigbrain = siibra.get_template('bigbrain').fetch(resolution_mm=res)
plotting.view_img(img, bg_img=bigbrain, cmap=mp.get_colormap(), symmetric_cmap=False)
```

Also, fixes voxel dimension calculation during mergin fragments of nifti volumes. (This was not a problem most of the time since the voxel size of most fragmented nifti is 1 mm)


siibra/volumes/providers/neuroglancer.py
```diff
@@ -195,7 +223,10 @@ def _merge_fragments(self) -> nib.Nifti1Image:
```
dickscheid, line 223: I know there is no change here, but I am thinking whether we might introduce holes here in edge cases, e.g. if the affine does significant stretches. How do we know that the Xt, Yt, Zt span gaplessly across the target voxel space?

maybe we should at this stage replace this part with nilearn.image.resample, which certainly deals with the problem?

dickscheid: A good example, although a bit artificial, is in case that the affine would require a rescaling by a factor of 2. Then the Xt would still be the same number of voxels as the Xs, although we would need double as many to fill the target volume without gaps.

AhmetNSimsek: There is one important change, same in nifti: `bbox.transform(np.linalg.inv(result_affine)).shape` instead of  `bbox.transform(result_affine).shape`. I think this is the correct way.

AhmetNSimsek: `nilearn` will certainly be more robust. However, to be sure we are talking about the same part, could you highlight or comment the lines? (214-220?)

dickscheid: I am referring to line 222 - does it always fill all target voxels? Without testing, by looking at the code, my feeling is that we take the nonzero voxel coordinates of the input, transform the coordinates to the resulting array, and fill in the input array voxel values at the target positions. If the affine spreads the space, my feeling is that the transformed input array voxels will not any more be gapless, e.g. map to disconnected voxels in the result image. Could you check this? If we used `nilearn.resample_to_img` for the resampling, we wouldn't have to deal with this here.

AhmetNSimsek: Please see [PR](https://github.com/FZJ-INM1-BDA/siibra-python/pull/572)

PR #563, opened 2024-02-19
fix: write lock for warmup

xgui3783:
Since warmup writes files, it is potentially possible to have multiple process trying to warmup simultaneously. 

This PR prevents simultaneous write by using a FileLock (or SoftFileLock in windows).

### DIFF
--- a/siibra/retrieval/cache.py
+++ b/siibra/retrieval/cache.py
@@ -208,8 +208,8 @@ class Warmup:
                 return
             for f in return_val:
                 f()
-        
-        with Lock(f"{SIIBRA_CACHEDIR}/warmup.tmp"):
+
+        with Lock(CACHE.build_filename("lockfile", ".warmup")):
             with ThreadPoolExecutor(max_workers=max_workers) as ex:
                 for _ in siibra_tqdm(
                     ex.map(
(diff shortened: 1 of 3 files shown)


siibra/retrieval/cache.py
```diff
@@ -23,6 +23,12 @@
 from typing import Callable, List, NamedTuple, Union
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
+import platform
+
+if platform.system() == "Linux":
+    from filelock import FileLock as Lock
+else:
+    from filelock import SoftFileLock as Lock
```
AhmetNSimsek, line 31: I think we can just use FileLock regardless: https://py-filelock.readthedocs.io/en/latest/index.html#filelock-vs-softfilelock

AhmetNSimsek (approved):
Just one comment, otherwise LGTM

siibra/retrieval/cache.py
xgui3783: copied directly from https://github.com/FZJ-INM1-BDA/siibra-python/blob/main/siibra/retrieval/requests.py#L47-L50

I thought there was an issue with FileLock on windows?

AhmetNSimsek: I can't find what could be the reason. It seems that is also your code; do you recall anything about it?

xgui3783: yea, me neither. 

Do you want me fix both instances?

AhmetNSimsek: Perhaps previously you saw the old api doc? https://github.com/tox-dev/filelock/issues/132

PR #565, opened 2024-02-23
fix: volume._pointsinside resample img to template

AhmetNSimsek:
During the intersection calculation of a volume and a pointset, `Volume._points_inside` determines the points laying inside the volume. However, some regional masks can have different shapes than the template. This can cause points to be extracted wrongly. At the moment, this only occurs during BigBrain intensity profile queries with a region that is defined on Julich Brain 2.9 on BigBrain.

To reproduce the issue:
```python
import matplotlib.pyplot as plt
import siibra
from nilearn import plotting

mask29 = siibra.get_region('julich 2.9', 'hoc1 left').get_regional_map('bigbrain').fetch()
tmp = siibra.get_template('bigbrain').fetch(resolution_mm=max(mask29.header.get_zooms()))
cf = siibra.features.get(
    siibra.get_region('julich 2.9', 'hoc1 left'),
    'BigBrainIntensityProfile'
)[0]
pts = cf.anchor.location.as_list()
print(len(pts))
display = plotting.plot_roi(mask29, bg_img=tmp)
display.add_markers(pts, marker_size=5)
plt.show()
```
(to test that it works otherwise, use `cf = siibra.features.get(siibra.get_region('julich 3', 'hoc1 left'), 'BigBrainIntensityProfile')[0]`)

TODO:
- [ ] add an example to test a possible scenario
- [ ] Update e2e tests

### DIFF
--- a/siibra/volumes/volume.py
+++ b/siibra/volumes/volume.py
@@ -220,11 +220,17 @@ class Volume(location.Location):
         """
         if not self.provides_image:
             raise NotImplementedError("Filtering of points by pure mesh volumes not yet implemented.")
-        img = self.fetch(format='image', **kwargs)
-        arr = img.get_fdata()
+        orgimg = self.fetch(format='image', **kwargs)
+        tmp = self.space.get_template().fetch(resolution_mm=max(orgimg.header.get_zooms()))
+        arr = resample_array_to_array(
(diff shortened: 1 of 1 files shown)


AhmetNSimsek:
This PR, in fact, does not solve the actual source of the bug. #566 solves the issue. The example here will be committed there.

PR #566, opened 2024-02-23
Add Volume.evaluate_points() and fix voxel access for _points_inside

dickscheid:


siibra/volumes/volume.py
```diff
@@ -214,29 +214,52 @@ def __repr__(self):
...
+        # fix the outside voxel values, which might have an inconsistent value now 
+        values[~inside] = outside_value
+
+        return values
+
+    def _points_inside(self, points: Union['point.Point', 'pointset.PointSet'], **kwargs) -> 'pointset.PointSet':
+        """
+        Reduce a pointset to the points which fall
+        inside nonzero pixels of this volume.
+        The indices of the original points are stored as point labels in the result.
```
dickscheid, line 257: in fact the given points might have labels already, which the user doesn't want to be overwritten by the indices in the input pointset. So I suggest to introduce an argument `return_original_indices_as_labels=True` which could be disabled. Or maybe better `keep_point_labels=False` which, when set to True, preserves the  labels of the input pointset instead of storing the point indices there. But please keep the behavior, it is used I think for the point clustering (could be checked)

siibra/volumes/volume.py
```diff
@@ -422,7 +445,7 @@ def draw_samples(self, N: int, sample_size: int = 100, e: float = 1, sigma_mm=No
             space=None
         )
         result = voxels.transform(img.affine, space='mni152')
-        result.sigma_mm = sigma_mm
+        result.sigma_mm = [sigma_mm for _ in result]
```
dickscheid, line 448: This has different reasons - we had in some cases a single float as the sigma for PointSet's, which had bad implications, so I wanted to make sure the sigma_mm in pointset is always a list. Please keep  this, better even introduce more formal ways that the sigma_mm is not overwritten to a single float

siibra/volumes/volume.py
AhmetNSimsek: Perhaps instead of altering the input points, we could return a list of indices that we use to label as well?
The user can just as easily label them from the output (because `PointSet.labels` is not a property). What do you think?

dickscheid: I think you are right. I support this idea.

AhmetNSimsek: As I implemented it, I changed my mind on this because we need to get the indices of the original PointSet after `intersection` call and not necessarily after `_points_inside` call.
So I think we should keep the current behavior since it creates a new PointSet anyway. (That is, while the return value is a reduced version of the original PointSet, it is still a new PointSet)

AhmetNSimsek (commented):
I made some comments and two commits. Could you take a look?

siibra/locations/pointset.py
```diff
@@ -34,11 +35,46 @@
     )
 
 
+def from_points(points: List["point.Point"], newlabels: List[int] = None) -> "PointSet":
```
dickscheid, line 38: Why wouldn't we just extend the PointSet constructor for this only slightly different variant?

siibra/locations/pointset.py
```diff
@@ -34,11 +35,46 @@
...
+
+
 class PointSet(location.Location):
     """A set of 3D points in the same reference space,
     defined by a list of coordinates."""
 
-    def __init__(self, coordinates, space=None, sigma_mm=0, labels: list = None):
+    def __init__(
+        self,
+        coordinates: Union[List[Tuple], np.ndarray],
```
dickscheid, line 73: if we allow here a list of Point in addition to list of Tuple, we can make the space optional.

Not fully sure if this is more elegant, but to me it seems a bit confusing to have from_points although it is almost identical to the coordinate-based constructor

siibra/locations/pointset.py
```diff
@@ -90,11 +129,16 @@ def intersection(self, other: location.Location):
         return intersection[0] if len(intersection) == 1 else intersection
 
     @property
-    def sigma(self):
-        return [p.sigma for p in self]
+    def coordinates(self) -> np.ndarray:
```
dickscheid, line 132: is this only to protect modification of the coordinates?

Python is usually not too restrictive with attributes. Was there an issue leaving the coordinates editable? should be mentioned here in a comment.

dickscheid: probably this makes sense, since modification of coordinates would have to maintain the sigma property as well.

siibra/volumes/volume.py
```diff
@@ -214,33 +214,107 @@ def __repr__(self):
...
+        SpaceWarpingFailedError
+            If warping of the points fails.
         """
         if not self.provides_image:
             raise NotImplementedError("Filtering of points by pure mesh volumes not yet implemented.")
-        img = self.fetch(format='image', **kwargs)
+
+        # make sure the points are in the same physical space as this volume
+        if isinstance(points, point.Point):  # requires to be a PointSet from next step onwards
+            as_pointset = pointset.from_points([points])
```
dickscheid, line 256: I think I've seen a similar thing at other places.

I would find it more elegant to implement Point.as_pointlist() instead.
This would indicate that we are semantically casting to the list type rather than creating a copy. 
What do you think?

siibra/volumes/volume.py
```diff
@@ -214,33 +214,107 @@ def __repr__(self):
...
+
+        # get the voxel array of this volume
+        img = self.fetch(format='image', **fetch_kwargs)
         arr = img.get_fdata()
-        warped = points.warp(self.space)
-        assert warped is not None
+
+        # transform the points to the voxel space of the volume for extracting values
         phys2vox = np.linalg.inv(img.affine)
         voxels = warped.transform(phys2vox, space=None)
```
dickscheid, line 268: I don't want to change anything here, but I repeatedly got across this space=None. 
I think it is important not to have None as default so that the user realizes that we create a location outside the usual reference spaces.
However, the resulting space is not none, it is the pixel space of the image. In the future, we might want to create an id for the pixel space, and put that here.
This should go into a feature request for future discussion. no changes in this PR

siibra/volumes/volume.py
```diff
@@ -214,33 +214,107 @@ def __repr__(self):
...
-            XYZ[np.all((XYZ < arr.shape) & (XYZ > 0), axis=1), :],
-            3, axis=1
-        )
-        arr[0, 0, 0] = 0  # ensure the lower left voxel is not foreground
-        inside = np.where(arr[X, Y, Z] != 0)[0]
-        return pointset.PointSet(
-            points.homogeneous[inside, :3],
-            space=points.space,
-            labels=inside
+        XYZ = voxels.coordinates.astype('int')
```
dickscheid, line 269: I didn't test, but I thought the `.homogeneous` output was 4xN to prepare it for matrix multiplication, while  `.coordinates` is Nx3 (and I would expect it as such). Can you please double-check that this consistent, and handled appropriately? 
The old code suggests that homogeneous was also Nx4, but I am a bit surprised and unsure, we should verify again.

siibra/volumes/volume.py
```diff
@@ -214,33 +214,107 @@ def __repr__(self):
...
+
+        Returns
+        -------
+        PointSet
+            A new PointSet containing only the points inside the volume.
+            Labels reflect the indices of the original points if `keep_labels`
+            is False.
+        """
+        values = self.evaluate_points(points, **fetch_kwargs)
+        inside = list(np.where(values != 0)[0])
```
dickscheid, line 312: didn't we want to introduce a "background" parameter with default 0 instead of hardcoding the zero?

siibra/locations/pointset.py
AhmetNSimsek: It was previously possible to do so but during the last refactoring, if I remember correctly, this was removed. I wrote this method to continue with the same logic of `volume.from_something()`.
Also, when the name of the keyword is `coordinates`, it is not elegant to supply points that have more information.

siibra/locations/pointset.py
AhmetNSimsek: please see https://github.com/FZJ-INM1-BDA/siibra-python/pull/566#discussion_r1518920209

siibra/locations/pointset.py
AhmetNSimsek: In the change log, it seems like sigma is edited out for coordinates but sigma has just moved below as a new property called `coordinates` added. The reason is that the `coordinates` used to be an attribute that is exposed and changeable, which potentially changes the data and makes a new point/pointset.

siibra/volumes/volume.py
AhmetNSimsek: I thought about such a method as well but while PointSet inherits Points, the vice versa is not true. I am not sure it is a good design to create this cyclical dependency (especially since Point class should exist solely as it is a building block of PointSet IMO). Also, I do not think `as_pointlist()` makes sense for a single point and I am not sure if there is use for users besides internal implementation.

siibra/volumes/volume.py
AhmetNSimsek: This is sth Xiao and I also have discussed; voxel space as a space should be possible. (Specifically, VoxelSpace under space.py. This would then point to a coordinate space still but requires information on how to get there, i.e. affine etc.) We plan to discuss this indeed

siibra/volumes/volume.py
AhmetNSimsek: We already did for `evaluate_points`. And I think we didn't implement for _points_inside as it is an internal method and in siibra we internally assume the background is 0. I can add the change but than user won't have access to it or we need to add this keyword to methods like intersection and assign.

AhmetNSimsek:
I think we should also consider using `evaluate_points` instead of `parcelationmap.Map._read_voxel`

siibra/volumes/volume.py
AhmetNSimsek: I checked every reference where `homogeneous is used. If it is used for matrix multiplication, it is transposed. So it is consistent.
There are a few redundant uses though, such as `XYZ = self.homogeneous[:, :3]` in `PointSet.boundingbox`. This could just be `XYZ = self.coordinates` as now it consistently produces a Nx3 `nd.array`.

PR #567, opened 2024-02-24
Exclude background from connected_components

dickscheid:
The `commons.connected_components` always returned at least two components - the background and the actual foreground structure. Of course the background is not desired. This PR excludes it.

I am quite sure that earlier siibra versions did not include the background component.

### DIFF
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -534,11 +534,12 @@ def connected_components(imgdata: np.ndarray):
     Provide an iterator over connected components in the array
     """
     from skimage import measure
-    components = measure.label(imgdata > 0)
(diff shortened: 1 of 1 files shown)


AhmetNSimsek (approved):
Yes, you are right. See https://github.com/FZJ-INM1-BDA/siibra-python/blob/39f22687fb0464a4251ad121edf75af41f24821f/siibra/commons.py#L422 for version 0.4

Just one note: Perhaps we can use `background` kwarg:
```python
components = measure.label(imgdata, background=0, connectivity=2)
```

dickscheid:
background kwarg is ok, I had this idea as well but wanted to keep the change minimal. If you introduce it, fine.

PR #572, opened 2024-02-28
Use `nilearn.resample_to_img` for mreging fragments

AhmetNSimsek:
Based on the [conversation ](https://github.com/FZJ-INM1-BDA/siibra-python/pull/561/files#r1496115337) at #561 
Code to test:
```python
import siibra
from nilearn import plotting
res = 0.68
fmt = 'neuroglancer/precomputed'
bg_img = siibra.get_template('mni152').fetch(format=fmt, resolution_mm=res)
mp = siibra.get_map('2.9', 'mni152')
cmap = mp.get_colormap()
img = mp.fetch(format=fmt, resolution_mm=res)
plotting.view_img(img, bg_img=bg_img, symmetric_cmap=False, cmap=cmap).open_in_browser()
```

### DIFF
--- a/siibra/commons.py
+++ b/siibra/commons.py
@@ -519,8 +519,7 @@ def resample_array_to_array(
     """
     from nibabel import Nifti1Image
     from nilearn.image import resample_to_img
-    interp = "nearest" if issubclass(source_data.dtype.type, np.integer) \
-        else 'linear'
+    interp = "nearest" if issubclass(source_data.dtype.type, np.integer) else 'linear'
     resampled_img = resample_to_img(
         Nifti1Image(source_data, source_affine),
         Nifti1Image(target_data, target_affine),
--- a/siibra/volumes/providers/neuroglancer.py
+++ b/siibra/volumes/providers/neuroglancer.py
@@ -88,7 +88,7 @@ class NeuroglancerProvider(_provider.VolumeProvider, srctype="neuroglancer/preco
             if fragment is None:
                 logger.info(
                     f"Merging fragments [{', '.join(self._fragments.keys())}]. "
-                    f"You can select one using `fragment` kwarg.)."
+                    f"You can select one using `fragment` kwarg."
                 )
                 result = self._merge_fragments(resolution_mm=resolution_mm, voi=voi, **kwargs)
             else:
@@ -178,7 +178,11 @@ class NeuroglancerProvider(_provider.VolumeProvider, srctype="neuroglancer/preco
         voi: _boundingbox.BoundingBox = None,
         **kwargs
     ) -> nib.Nifti1Image:
+        from nilearn.image import resample_to_img
+
         # TODO this only performs nearest neighbor interpolation, optimized for float types.
+        interp = lambda tp: "nearest" if issubclass(tp, np.integer) else 'linear'
+
         with QUIET:
             bbox = self.get_boundingbox(
                 clip=False,
@@ -186,6 +190,7 @@ class NeuroglancerProvider(_provider.VolumeProvider, srctype="neuroglancer/preco
                 resolution_mm=resolution_mm,
                 voi=voi
             )
+
         num_conflicts = 0
         result = None
         for frag_vol in self._fragments.values():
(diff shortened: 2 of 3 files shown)


siibra/volumes/providers/nifti.py
```diff
@@ -108,7 +108,11 @@ def get_boundingbox(self, clip=True, background=0, **fetch_kwargs) -> "_bounding
         return bbox
 
     def _merge_fragments(self) -> nib.Nifti1Image:
+        from nilearn.image import resample_to_img
+
         # TODO this only performs nearest neighbor interpolation, optimized for float types.
+        interp = lambda tp: "nearest" if issubclass(tp, np.integer) else 'linear'
```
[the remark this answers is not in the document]
dickscheid: as an additional comment, lambda functions are typically considered bad style except used where they are defined. I find it ok personally to use for such purpose, but would move it as close as possible to where it is used (if used only once, I would put it there right away)

dickscheid (commented):
How is the performance compared to the previous implementation? comparable?

siibra/volumes/providers/neuroglancer.py
```diff
@@ -178,14 +178,19 @@ def _merge_fragments(
         voi: _boundingbox.BoundingBox = None,
         **kwargs
     ) -> nib.Nifti1Image:
+        from nilearn.image import resample_to_img
+
         # TODO this only performs nearest neighbor interpolation, optimized for float types.
+        interp = lambda tp: "nearest" if issubclass(tp, np.integer) else 'linear'
```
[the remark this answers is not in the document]
AhmetNSimsek: Hmm. I don't think adding another parameter is a good idea because we are introducing merging fragments in case the user requests a map without specifying fragments, which usually occurs by calling `map.fetch()` without any arguments. And using this parameter would require the implicit knowledge of the image content.

However, I see what you mean. I will see if I can find a better solution.

siibra/volumes/providers/nifti.py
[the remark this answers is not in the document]
AhmetNSimsek: > same as above.
> 
> Since we are implementing this twice now, it might be worth checking if this can be merged to a parent class method partially.

I considered the same thing. We have a method in commons for this. But let us first agree on the usage and performance.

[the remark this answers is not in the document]
AhmetNSimsek: > as an additional comment, lambda functions are typically considered bad style except used where they are defined. I find it ok personally to use for such purpose, but would move it as close as possible to where it is used (if used only once, I would put it there right away)

I prefer not to use them but we agreed before that it is not sth we track in terms of bad style.

As it is a small function, I think it is easy to find. Putting it where it is used makes the line repeated for no good reason. However, I can make the change if you prefer.

siibra/commons.py
```diff
@@ -519,12 +519,10 @@ def resample_array_to_array(
     """
     from nibabel import Nifti1Image
```
dickscheid, line 520: I am not quite sure why we put this into commons. The function call is not shorter or simpler than the actual implementation (ie the call to the nilearn function), we just forward the arguments here. Shouldn't we just use the nilearn function on the calling side ride away (in neuroglancer.py)? especially since we import the nilearn package locally only...

siibra/volumes/providers/neuroglancer.py
```diff
@@ -200,32 +207,28 @@ def _merge_fragments(
...
-                    np.linalg.inv(result_affine),
-                    np.dot(img.affine, np.c_[Xs, Ys, Zs, Zs * 0 + 1].T)
-                ) + .5).astype('int'),
-                4, axis=0
+            # resample to merge template and update it
+            arr = resample_array_to_array(
+                source_data=np.asanyarray(img.dataobj),
+                source_affine=img.affine,
+                target_data=result_arr,
+                target_affine=result_affine
```
dickscheid, line 221: See above, this is more or less 1:1 doing the call to nilearn.image.resample_to_img with the same arguments, except for casting to Niftti1Image, but we have Nifti1Image imported here already so could call it right away, no?

siibra/volumes/providers/neuroglancer.py
```diff
@@ -200,32 +207,28 @@ def _merge_fragments(
...
             )
-            num_conflicts += np.count_nonzero(result_arr[Xt, Yt, Zt])
-            result_arr[Xt, Yt, Zt] = arr[Xs, Ys, Zs]
+            nonzero_voxels = arr != 0
+            num_conflicts += np.count_nonzero(result_arr[nonzero_voxels])
```
dickscheid, line 224: a possible advantage of using the X Y z coordinate lists over this is that for sparse volumes (such as probability maps), the coordinate list is much smaller in memory compared to the full boolean volume, so I think we might save a lot of working memory potentially by using np.where() and passing the  the coordinates. I never tested this, though - would be interesting to profile the two possibilities in terms of speed and memory.

siibra/commons.py
AhmetNSimsek: I think it would be better indeed! At the time, I think you wanted to remove the package import in volumes, which has a valid point but I think it doesn't make much of a difference. If we do this tho, we will need to change a few uses where we specifically convert the image to array etc. But I think it is more robust and clear.

Tho I am not against the idea of having a function in NiftiProvider that wraps `resample_to_img` with our preferrd interpolation choice so that we don't repeat the line in several places.

siibra/volumes/providers/neuroglancer.py
AhmetNSimsek: I totally agree. Please see above

siibra/volumes/providers/neuroglancer.py
AhmetNSimsek: We don't merge probability maps (and IMO we must not) so I am not sure this is worth pursuing.

PR #573, opened 2024-02-28
Documentation for architecture decision records

AhmetNSimsek:
Finding architectural decisions and options considered in old meeting notes is difficult. This PR proposes storing the decisions under documentation as a part of developer documentation (an industry standard).


AhmetNSimsek:
In the Dev meeting on 29.02.2024, it is decided that this is not necessary.

PR #577, opened 2024-03-10
fix: pointset intersection with boundingbox fails if no points inside

dickscheid:
The pointset - boundingbox on main raised an exception if no points were inside the bonding box.

This PR should fix the problem.


AhmetNSimsek (approved):
One point but it is not a problem to have it there. LGTM

PR #582, opened 2024-03-14
feat: update how related regions are fetched

xgui3783:
see https://github.com/FZJ-INM1-BDA/siibra-configurations/pull/44 

This PR is the companion PR for updating the SANDS reference within siibra. 

Rather than relying on SANDS references of parcellation entity version, it is updated to use parcellation entity, which semantically more closely models SANDS model.

n.b. as this PR requires a specific branch of siibra configuration to function properly, this PR also aims to introduce the mechanism, by which, based on commit message, custom ref of siibra-configuration will be used. 

These changes are not yet pushed, in order to follow the Red/Green light approach for TDD


xgui3783:
Expecting tests to fail

AhmetNSimsek (changes requested):
LGTM.
Could you add docstrings to `RegionRelationAssessments` members?

AhmetNSimsek (approved):
Thanks, looks good.

For future reference, would you mind using numpy style as in https://github.com/FZJ-INM1-BDA/siibra-python/blob/50d4d4c3db1f185a0666cd793127748c6a904cb3/siibra/core/region.py#L273-L297
This should also work but better to have a unified style. Please don't waste time on it now tho. There are more urgent things atm.

xgui3783:
> Thanks, looks good.
> 
> For future reference, would you mind using numpy style as in
> 
> https://github.com/FZJ-INM1-BDA/siibra-python/blob/50d4d4c3db1f185a0666cd793127748c6a904cb3/siibra/core/region.py#L273-L297
> 
> 
> This should also work but better to have a unified style. Please don't waste time on it now tho. There are more urgent things atm.

Apologies, I have a lot on my plate right now. 

Do you mind taking over the reformatting for me?

@AhmetNSimsek

AhmetNSimsek:
Since the config branch had been merged and tagged, bumped the version ([3b2d9fa](https://github.com/FZJ-INM1-BDA/siibra-python/pull/582/commits/3b2d9fae7da120818e810f086bf84dc291ed492b)) to test with the correct config.

PR #583, opened 2024-03-25
Use precalculated image feature boundingboxes to speed up image feature query

AhmetNSimsek:
The profiling revealed that most time is spend fetching the boundingboxes of images. When this information is cached, the query itself takes less than a second (on 11th Gen Intel(R) Core(TM) i5-1145G7, 16 GB memory). Therefore, this PR allows the bounding box information to be precalculated and stored in the configs to be used in run time. I suggest that we utilize this method for all image features until the new spatial query system is settled.

### DIFF
--- a/siibra/features/image/image.py
+++ b/siibra/features/image/image.py
@@ -21,6 +21,7 @@ from .. import anchor as _anchor
 
 from ...volumes import volume as _volume
 from ...volumes.providers import provider
+from ...locations import boundingbox
 
 from typing import List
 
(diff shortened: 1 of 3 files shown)


siibra/features/image/image.py
```diff
@@ -83,6 +85,15 @@ def __init__(
...
         self._name_cached = name
+        self._boundingbox = boundingbox
+
+    def get_boundingbox(
+        self, clip: bool = True, background: float = 0.0, **fetch_kwargs
+    ) -> "boundingbox.BoundingBox":
+        if self._boundingbox is None:
+            return super().get_boundingbox(clip, background, **fetch_kwargs)
+        else:
+            return self._boundingbox
```
AhmetNSimsek, line 96: This could be in volume to have the flexibility, i.e., the non-feature volumes (templates, probability maps...) can have recalculated bounding boxes as well.

dickscheid (approved):
This is a good idea.

PR #589, opened 2024-04-18
fix: memory usage when warming the cache

xgui3783:
Current implementation of warming data results in memleak (well, we call it in memory caching)

this PR fixes this, and the memory foot print is consistently < 1G when warming the cache. For comparison, without this PR, memory footprint goes beyond 4G, when the process is OOM killed.


AhmetNSimsek (approved):
Good catch! Thanks!

xgui3783:
it looks like `MRIVolumeOfInterest` etc does not have plot implemented, thus the e2e failing. 

Should we remove them from the plot testing list?

@AhmetNSimsek

PR #592, opened 2024-04-23
Test examples as a part of e2e. Create docs on push/PR to main and on publish

AhmetNSimsek:


.github/workflows/docs.yml
```diff
@@ -1,7 +1,9 @@
 name: '[Docs and examples] Create docs and run examples'
-on: 
-  pull_request:
+on:
+  push:
     branches: [ 'main' ]
+  pull_request:
+    branches: [ 'main' ]
```
xgui3783, line 6: I would argue

```
on:
  push: branches: [ "**" ]
```

and remove the `pull_request` trigger

e2e/test_examples.py
```diff
@@ -0,0 +1,16 @@
+import pathlib
```
xgui3783, line 1: see remark

xgui3783 (commented):
since `.github/workflows/docs.yml` also runs examples, IMO, it makes more sense to run examples in `.github/workflows/docs.yml` exclusively. 

We could rename the workflow as `examples-and-build-doc`

AhmetNSimsek:
> since `.github/workflows/docs.yml` also runs examples, IMO, it makes more sense to run examples in `.github/workflows/docs.yml` exclusively.
> 
> We could rename the workflow as `examples-and-build-doc`

But then artifacts will be uploaded for no reason and if sphinx-build fails due to sphinx/documentation-related issues, the examples won't be passing.

e2e/test_examples.py
```diff
@@ -0,0 +1,16 @@
...
+import runpy
+import pytest
+import sys
+
+examples = pathlib.Path('./examples').resolve().rglob('*.py')
+python_version = sys.version_info
+
+
+@pytest.mark.skipif(
+    (python_version.major, python_version.minor) == (3, 7),
```
AhmetNSimsek, line 11: should be "(python_version.major, python_version.minor) != (3, 7)"

AhmetNSimsek: After a discussion with @xgui3783 on 26.04.2024, we decided to test examples as part of e2e full or fast workflows. So the tests should not be skipped and python version is determined by e2e workflows.

PR #594, opened 2024-05-14
refactor: use attributes

xgui3783:
supercedes https://github.com/FZJ-INM1-BDA/siibra-python/pull/553


xgui3783:
attn @AhmetNSimsek 

per our discussion, here is the WIP

left to do:

refactor todo list

- [ ] update `siibra/attr/meta/doi` -> `siibra/attr/meta/url`
    - [ ] update schema
    - [ ] update key `url` -> `value`
- [ ] update `siibra/attr/data/voi` -> `siibra/attr/data/image`
    - [ ] update schema
- [ ] update `regionspec`
    - [ ] fix property key `name` -> `value`
- [ ] reimplement autocomplete all modalities
- [ ] reimplement matching between attributes
    - [ ] regionspec match
- [ ] move cache to root level module 
- [ ] add more attr comparison
- [ ] add test for attr comparison
- [ ] reimplement volume providers?
- [ ] consider using standard json-ld schema, context etc

AhmetNSimsek:
`region.find` is not deterministic and rather slow. I am noting down to investigate later.

AhmetNSimsek:
@xgui3783, I think we should remove the `dataclass` decorators from classes that inherit other a class that has a `dataclass` decorator. Because it creates unexpected behaviors such as some methods are overridden and very difficult to debug as it is an unexpected behavior.
Regarding being explicit: if a class `A` inherits a class `B` with `dataclass` decorators, the developer would know that `A` is also a `dataclass`. So, by repeating the 2 lines required to decorate class `A` as a `dataclass`, we break DRY, we face unexpected behavior, and we don't gain anything besides being explicit which is not really necessary in this context.

siibra/retrieval_new/api_fetcher/brainglobe.py
```diff
@@ -107,13 +116,33 @@ def populate_regions(
         parent_region = _dict_id_to_region[parent_id]
         region.parent = parent_region
 
+def tiff_to_nii(tiff_bytes: bytes, affine: np.ndarray) -> str:
```
AhmetNSimsek, line 119: I think this should go to commons_new.maps or we create a tiff fetcher.
I prefer the second one

siibra/retrieval_new/api_fetcher/brainglobe.py
```diff
@@ -138,20 +173,42 @@ def bg_parcellation():
...
+    
+    @attribute_collection_iterator.register(Map)
+    def bg_map():
+        return [labelled_map]
 
     _registered_atlas.add(atlas_name)
     logger.info(f"{atlas_name} added to siibra.")
 
-    return space, parcellation
+    return space, parcellation, labelled_map
```
AhmetNSimsek, line 214: Just food for discussion: since a labelled_map has to have this space and parcellation as a member, or computed property, perhaps we should just return the map.
The user can then access to parcellation and space either from `mp.parcellation/space` or from the registry

xgui3783: I don't mind either way (I am also fine not returning anything)

The only reason why I return all three is to be explicit, to show all of the items that are added

siibra/retrieval_new/api_fetcher/brainglobe.py
xgui3783: as you alluded to, there is no standard tiff format. 

That's why I think it should be here (i.e. brainglobe standard)

AhmetNSimsek:
To profile and improve:
- [ ] `get_map` is very slow. It does not speed up after first use.
- [ ]

Known bugs:
- [ ]

Other issues:(
- [ ] `get_parcellation` does not return the newest parcellation

@xgui3783 Please edit and add/update more as we move forward to keep track of finer details

PR #603, opened 2024-09-23
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
Again, great talk at INCF 2024 @dickscheid !  Amazing scale of endeavors.  Need to try siibra-python and see how we could integrate your online options with our external servicing for https://dandiarchive.org/ .  Let's chat more while in the same room ;)

Meanwhile - my humble contribution to siibra is this PR.
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.css,.codespellrc
+check-hidden = true
+ignore-regex = ^\s*"image/\S+": ".*
+# ignore-words-list =
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,8 @@
 [codespell]
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
-skip = .git*,*.css,.codespellrc
+# TODO: go through gene_names.json manually since might have some true positives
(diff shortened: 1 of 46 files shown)


xgui3783:
Dear @yarikoptic ,

Thank you for your interest in siibra-python. 

We are in the middle of a large refactor at the moment. I would imagine your PR would almost certainly cause merge conflicts. 

We understand your time is precious, and would thus like to offer to implement codespell once refactor is stable, and attribute you as a co-contributor[1].

If you would prefer to create new PR/commits once refactor becomes stable, we can also offer to ping you when that is the case (if this is the case, please let us know in this issue). 

In either case, we will keep this PR open until it is resolved in one of the two ways above.

[1] https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

yarikoptic:
Whatever works for you. Ping be whenever

yarikoptic:
coolio.
ping me whenever you want/ready for me to rebase/rerun fixup of typos so we avoid conflicts (unless you want to just merge and resolve yourself).

AhmetNSimsek:
Hi @yarikoptic Would you mind creating a new branch from the current main and create a new PR?

PR #609, opened 2024-10-24
fix bbox planar intersection

AhmetNSimsek:
This PR fixes the no intersection issue when the intersection of 2 bounding boxes is planar and hence have volume 0 resulting in None. This meant that while there are some intersections, these were ignored. Also, the edge case, the intersection is just a point, is accounted for.

code to reproduce the issue on [current main](31b091200d7c3f3d041e762925ff6571d862268a)
```python
import siibra
points = siibra.PointSet(
    [
        [-35.72, -1.87, 35.03],
        [-33.7, -1.87, 36.2],
        [-32.56, -1.87, 34.34],
        [-35.62, -1.87, 35.2],
    ],
    "bigbrain",
)
matched_sections = siibra.features.get(
    points, siibra.features.cellular.CellbodyStainedSection
)
assert len(matched_sections) > 0
voi = points.boundingbox

# this is expected to fetch but fails assuming there is no intersection
patch_1mu = matched_sections[0].fetch(voi=voi, resolution_mm=-1)
```

### DIFF
--- a/siibra/locations/boundingbox.py
+++ b/siibra/locations/boundingbox.py
@@ -115,12 +115,12 @@ class BoundingBox(location.Location):
     def __str__(self):
         if self.space is None:
             return (
-                f"Bounding box from ({','.join(f'{v:.2f}' for v in self.minpoint)}) mm "
-                f"to ({','.join(f'{v:.2f}' for v in self.maxpoint)}) mm"
+                f"Bounding box from ({','.join(f'{v:.2f}' for v in self.minpoint)})mm "
+                f"to ({','.join(f'{v:.2f}' for v in self.maxpoint)})mm"
             )
         else:
             return (
-                f"Bounding box from ({','.join(f'{v:.2f}' for v in self.minpoint)}) mm "
+                f"Bounding box from ({','.join(f'{v:.2f}' for v in self.minpoint)})mm "
                 f"to ({','.join(f'{v:.2f}' for v in self.maxpoint)})mm in {self.space.name} space"
             )
 
--- a/siibra/locations/boundingbox.py
+++ b/siibra/locations/boundingbox.py
@@ -297,9 +297,10 @@ class BoundingBox(location.Location):
             return self
         else:
             try:
-                return self.corners.warp(spaceobj).boundingbox
-            except ValueError:
+                warped_corners = self.corners.warp(spaceobj)
+            except SpaceWarpingFailedError:
                 raise SpaceWarpingFailedError(f"Warping {str(self)} to {spaceobj.name} not successful.")
+            return warped_corners.boundingbox
 
     def transform(self, affine: np.ndarray, space=None):
         """Returns a new bounding box obtained by transforming the
--- a/siibra/locations/boundingbox.py
+++ b/siibra/locations/boundingbox.py
@@ -188,12 +188,18 @@ class BoundingBox(location.Location):
                 result_minpt.append(A[dim])
                 result_maxpt.append(B[dim])
 
+        if result_minpt == result_maxpt:
+            return result_minpt
+
         bbox = BoundingBox(
             point1=point.Point(result_minpt, self.space),
             point2=point.Point(result_maxpt, self.space),
             space=self.space,
         )
-        return bbox if bbox.volume > 0 else None
+
+        if bbox.volume == 0 and sum(cmin == cmax for cmin, cmax in zip(result_minpt, result_maxpt)) == 2:
+            return None
+        return bbox
 
     def _intersect_mask(self, mask: 'Nifti1Image', threshold=0):
         """Intersect this bounding box with an image mask. Returns None if they do not intersect.
--- a/siibra/locations/point.py
+++ b/siibra/locations/point.py
@@ -55,10 +55,14 @@ class Point(location.Location):
             if len(digits) == 3:
                 return tuple(float(d) for d in digits)
         elif isinstance(spec, (tuple, list)) and len(spec) in [3, 4]:
+            if any(v is None for v in spec):
+                raise RuntimeError("Cannot parse cooridantes containing None values.")
             if len(spec) == 4:
                 assert spec[3] == 1
             return tuple(float(v.item()) if isinstance(v, np.ndarray) else float(v) for v in spec[:3])
         elif isinstance(spec, np.ndarray) and spec.size == 3:
+            if any(np.isnan(v) for v in spec):
+                raise RuntimeError("Cannot parse cooridantes containing NaN values.")
             return tuple(float(v.item()) if isinstance(v, np.ndarray) else float(v) for v in spec[:3])
         elif isinstance(spec, Point):
             return spec.coordinate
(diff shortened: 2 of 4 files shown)


siibra/locations/boundingbox.py
```diff
@@ -188,12 +188,18 @@ def _intersect_bbox(self, other: 'BoundingBox', dims=[0, 1, 2]):
                 result_minpt.append(A[dim])
                 result_maxpt.append(B[dim])
 
+        if result_minpt == result_maxpt:
+            return result_minpt
```
AhmetNSimsek, line 192: This is not so important. Just an edge case which perhaps should not be included.

xgui3783:
LGTM, but you should check with @dickscheid to see if volume check is important. 

remind me, `restrict_space=False` is the default for `siibra.features.get()` right?

This issue does not affect siibra python v2, since there is no volume check.

AhmetNSimsek:
> LGTM, but you should check with @dickscheid to see if volume check is important.

There is a conversation about it [here](https://github.com/FZJ-INM1-BDA/siibra-python/pull/591#discussion_r1593404112). I thought I remembered all the details but I will make some adjustments to this PR based on Timo's comment.

> remind me, `restrict_space=False` is the default for `siibra.features.get()` right?

Yes by default it is false, i.e., compares anchors across different spaces.

> This issue does not affect siibra python v2, since there is no volume check.

True

AhmetNSimsek:
`NaN` centroids will be addressed in a separate PR since changes here only revealed them, not caused them.

dickscheid:
> ```python
> import siibra
> points = siibra.PointSet(
>     [
>         [-35.72, -1.87, 35.03],
>         [-33.7, -1.87, 36.2],
>         [-32.56, -1.87, 34.34],
>         [-35.62, -1.87, 35.2],
>     ],
>     "bigbrain",
> )
> matched_sections = siibra.features.get(
>     points, siibra.features.cellular.CellbodyStainedSection
> )

this should reveal sections where a subset of points is in side the section's nxmx0.02 mm bounding box.

> assert len(matched_sections) > 0
> voi = points.boundingbox

This should reveal a bounding box with zero volume.

> # this is expected to fetch but fails assuming there is no intersection
> patch_1mu = matched_sections[0].fetch(voi=voi, resolution_mm=-1)
> ```

This is not expected to fetch, since the voi is zero volume. The user would need to extend the voi in the z direction to at least 20 micrometer. I would expect an error message in this direction. There was a tutorial notebook at some point which did this, but we might have removed that part.

dickscheid:
Further to the above, I think in such cases a desired solution for users I to
1) determine the points which are truly inside the section
2) project them to the "walls" of the section's bounding box
3) obtain the bounding box of the resulting points.
But that is definitely a modification of the previous bounding box and should be modeled with an explicit function, not performed implicitly.

AhmetNSimsek:
> 
> > ```python
> > import siibra
> > points = siibra.PointSet(
> >     [
> >         [-35.72, -1.87, 35.03],
> >         [-33.7, -1.87, 36.2],
> >         [-32.56, -1.87, 34.34],
> >         [-35.62, -1.87, 35.2],
> >     ],
> >     "bigbrain",
> > )
> > matched_sections = siibra.features.get(
> >     points, siibra.features.cellular.CellbodyStainedSection
> > )
> > ```
> 
> this should reveal sections where a subset of points is in side the section's nxmx0.02 mm bounding box.
> 
> > assert len(matched_sections) > 0
> > voi = points.boundingbox
> 
> This should reveal a bounding box with zero volume.
> 
> > ```python
> > # this is expected to fetch but fails assuming there is no intersection
> > patch_1mu = matched_sections[0].fetch(voi=voi, resolution_mm=-1)
> > ```
> 
> This is not expected to fetch, since the voi is zero volume. The user would need to extend the voi in the z direction to at least 20 micrometer. I would expect an error message in this direction. There was a tutorial notebook at some point which did this, but we might have removed that part. 
> 
> 

The code is directly from the bigbrain workshop tutorial. Assuming we expect the tutorials to pass, the above code snippet is expected to fetch, is it not?

siibra/locations/boundingbox.py
dickscheid: the solutions makes a lot of sense

siibra/locations/boundingbox.py
```diff
@@ -188,12 +188,18 @@ def _intersect_bbox(self, other: 'BoundingBox', dims=[0, 1, 2]):
...
+
         bbox = BoundingBox(
             point1=point.Point(result_minpt, self.space),
             point2=point.Point(result_maxpt, self.space),
             space=self.space,
         )
-        return bbox if bbox.volume > 0 else None
+
+        if bbox.volume == 0 and sum(cmin == cmax for cmin, cmax in zip(result_minpt, result_maxpt)) == 2:
```
dickscheid, line 200: would be good to put a comment here, this is not trivial:
we check the three dimensions of the interesection bounding box. If exactly two dimensions are zero (the min and max point in these dimensions are identical), we conclude that there is no intersection.

dickscheid: This neglects the case where the intersection bounding box is planar, but has a sigma defined which gives it a nonzero "thickness" nevertheless.

@AhmetNSimsek how is the sigma computed from the original bounding boxes? we should discuss this. For now, I would be fine with the given solution but there should be a todo or even warning for bounding boxes with sigma. We should have a meeting on the processing of point uncertainties in the different location operations, for siibra v2. siibra v1 should try to raise warnings if locations have a sigma defined, but not used in operations.

siibra/locations/point.py
```diff
@@ -55,10 +56,14 @@ def parse(spec, unit="mm") -> Tuple[float, float, float]:
             if len(digits) == 3:
                 return tuple(float(d) for d in digits)
         elif isinstance(spec, (tuple, list)) and len(spec) in [3, 4]:
+            if any(v is None for v in spec):
+                raise RuntimeError("Cannot parse cooridantes containing None values.")
```
dickscheid, line 60: typo

siibra/locations/point.py
```diff
@@ -55,10 +56,14 @@ def parse(spec, unit="mm") -> Tuple[float, float, float]:
...
             if len(spec) == 4:
                 assert spec[3] == 1
             return tuple(float(v.item()) if isinstance(v, np.ndarray) else float(v) for v in spec[:3])
         elif isinstance(spec, np.ndarray) and spec.size == 3:
+            if any(np.isnan(v) for v in spec):
+                raise RuntimeError("Cannot parse cooridantes containing NaN values.")
```
dickscheid, line 66: typo

siibra/locations/point.py
```diff
@@ -137,9 +142,9 @@ def warp(self, space):
             z=self.coordinate[2],
         )
         response = HttpRequest(url, lambda b: json.loads(b.decode())).get()
-        if any(map(np.isnan, response['target_point'])):
-            logger.info(f'Warping {str(self)} to {spaceobj.name} resulted in NaN')
-            return None
+        if np.any(np.isnan(response['target_point'])):
+            raise SpaceWarpingFailedError(f'Warping {str(self)} to {spaceobj.name} resulted in NaN')
+
```
dickscheid, line 147: good idea to define an exception for this

Issue #157, opened 2022-02-14
[bug] spatial props for bundle atlas not calculated correctly


xgui3783:
as of v0.3a11

the spatial props calculated for regions such as lh_PoCi-PrCu_1 can be misleading, due to the concaved geometry of the probabilistic map:
<img width="507" alt="image002" src="https://user-images.githubusercontent.com/19381783/153879507-ef07e731-34f2-46c4-9a78-ad11c0511f8b.png">

dickscheid:
This is not suprising, since we just compute the mean. The mean is certainly not a very good choice for fibrre bundles - we might choose the closest point to the mean which is inside the structure.

AhmetNSimsek:
Alternative suggestion: ["Find the pixel with the highest closeness centrality. Closeness centrality is the inverse of the total sum of shortest distances from a node to every other node."](https://scikit-image.org/docs/stable/api/skimage.graph.html#skimage.graph.central_pixel) (see https://scikit-image.org/docs/stable/auto_examples/applications/plot_pixel_graphs.html)

```python
from time import time
import skimage as ski
import siibra
from nibabel import Nifti1Image
import numpy as np

def calculate_graph_center(img: Nifti1Image):
    mask_arr = np.asanyarray(img.dataobj > 0, dtype='uint8')
    skeleton = ski.morphology.skeletonize(mask_arr )
    g, nodes = ski.graph.pixel_graph(skeleton, connectivity=2)
    px, distances = ski.graph.central_pixel(g, nodes=nodes, shape=skeleton.shape, partition_size=100)
    return np.dot(img.affine, np.concatenate((px, [1])))[:3]

img = siibra.get_region("Superficial fibre bundles", "lh_PoCi-PrCu_1").get_regional_mask("mni 152").fetch()
s = time()
center = calculate_graph_center(img)
print(time() - s)
print(center)
```

![Image](https://github.com/user-attachments/assets/3619b185-05bb-46a4-b1be-0eeee82cec5e)

PR #610, opened 2024-10-29
maint: unify centroid computation

AhmetNSimsek:
This PR fixes https://github.com/FZJ-INM1-BDA/siibra-python/issues/613 and https://github.com/FZJ-INM1-BDA/siibra-python/pull/202 by using the robust centroid computation used in `Region.spatial_props()`. Now all centroid computation is done by finding the components of an image with `skimage.measure.label.()` and then finding the center of the non-zero values in the voxel space.

Also:
- `volume.ComponentSpatialProperties` is introduced in place of `region.SpatialPropCmpt`. 
- `region.SpatialProp` is removed since it is not used and was only introduced for typing.
- added tests
- Allow user to not split to components

Not backward compatible:
Instead of using `SpatialProps.components` now the result of `Volume.compute_spatial_props()` is a list of `ComponentSpatialProperties` per component. If the image is not split to components then it will return a single `ComponentSpatialProperties`.

Note: this PR does not address https://github.com/FZJ-INM1-BDA/siibra-python/issues/157. The root cause of this issue is the concaved geometry of the regions.


xgui3783 (approved):
LGTM

PR #614, opened 2024-11-04
`NeuroglancerScale.fetch()` return nii w dataobj of xyz coords

AhmetNSimsek:


AhmetNSimsek:
This question will not be addressed until version 2.

Issue #276, opened 2023-02-21
[bug] cannot get boundary_positions for hoc2 right


xgui3783:
hash: b881335f36ed06c302007c5ff7963af7426c4a4a

sample code:

```python
import siibra
region = siibra.get_region("2.9", "hoc2 right")
features = siibra.features.get(region, "CorticalProfile")
print([f.boundary_positions for f in features])
```

expected output list of boundary positions

actual output

```python
  File "/home/xgui3783/dev/projects/siibra-python/siibra/features/cellular/cell_density_profile.py", line 183, in boundary_positions
    self._boundary_positions[b] = self.depth_image[
  File "/home/xgui3783/dev/projects/siibra-python/siibra/features/cellular/cell_density_profile.py", line 149, in depth_image
    D = np.zeros((np.array(self.density_image.shape) * scale).astype("int") + 1)
  File "/home/xgui3783/dev/projects/siibra-python/siibra/features/cellular/cell_density_profile.py", line 198, in density_image
    bins=(np.array(self.layer_mask.shape) / pixel_size_micron + 0.5).astype(
  File "/home/xgui3783/dev/projects/siibra-python/siibra/features/cellular/cell_density_profile.py", line 138, in layer_mask
    self._layer_mask[Y, X] = layer
IndexError: index 2603 is out of bounds for axis 0 with size 2603
```

xgui3783:
snapshot kg file hash

thereby generate feature id from file hash

AhmetNSimsek:
Note:
- The number of features is not consistent in different queries.
- Running the same code twice lists all positions without an error

AhmetNSimsek:
CellDensityProfile (Segmented cell body density) anchored at Area hOc2 (V2, 18) with Point in BigBrain microscopic template (histology) [20.872669219970703,-58.29999923706055,-6.678797721862793] has the issue.
(By the second run the previous boundary positions are cached so it does not fail twice.)

Issue #509, opened 2023-11-17
Some CellDensityProfiles have invalid last boundary position


AhmetNSimsek:
If asked to get the data or plot, some CellDensityProfiles fail due to the last boundary position being just above 1.0 (eg  1.0000067544767959). This should be addressed either at the data level or siibra should handle these.

Some example ids:
```
dae2408a-95d0-4f63-ab37-f0756dff4720--07ee22d5b0824effaf90b0affc3d31aa
dae2408a-95d0-4f63-ab37-f0756dff4720--3411389421e087c6d398686a9e3c51d1
dae2408a-95d0-4f63-ab37-f0756dff4720--333a99f54a8116e04fa12d4b53fa9e7c
dae2408a-95d0-4f63-ab37-f0756dff4720--fd709c4fa7f3f7967e04065daee15799
16a71c7c-8940-4fc8-b2a7-d667933c2f2a--eb556c3f525337e305e68f71885325f3
f25c6728-2b53-4304-aa36-6165bbca3971--ee445f7352a3699813280ffd92207cb6
f25c6728-2b53-4304-aa36-6165bbca3971--0fbf3532f6d62f6f99f567a3c1102594
f25c6728-2b53-4304-aa36-6165bbca3971--aa9a723c9e15c02ea3f7fd02e374b3c4
f25c6728-2b53-4304-aa36-6165bbca3971--079376d3461c173c1d19139f556f3f71
```
Test with:
```python
import siibra
f = siibra.siibra.features._get_instance_by_id("dae2408a-95d0-4f63-ab37-f0756dff4720--07ee22d5b0824effaf90b0affc3d31aa")
f._check_sanity()
```

xgui3783:
ditto for `lq0::BigBrainIntensityProfile::p:minds/core/parcellationatlas/v1.0.0/94c1125b-b87e-45e4-901c-00daee7f2579-300::r:Area hOc1 (V1, 17, CalcS) left::4c05163cac01b560cddf9d0ae2b63c94`

AhmetNSimsek:
solved with #617

PR #617, opened 2024-11-12
fix: cell density profile

AhmetNSimsek:
Fixes https://github.com/FZJ-INM1-BDA/siibra-python/issues/276 and https://github.com/FZJ-INM1-BDA/siibra-python/issues/509


AhmetNSimsek:
Failing tests are due to Allen API so the PR will be merged.

PR #638, opened 2025-01-14
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
Recreation of
- https://github.com/FZJ-INM1-BDA/siibra-python/pull/603

against current main.  Likely the last of the kind ;-)


yarikoptic:
oops -- fixed up that inline ignore now. lovely to see all the testing CI! ;)

AhmetNSimsek:
Thanks @yarikoptic !

PR #641, opened 2025-02-05
Allow random cmap creation for `map.get_color_map`

AhmetNSimsek:
```python
import siibra
from nilearn import plotting
mp = siibra.get_map("deep bundles", "mni152")
cmap = mp.get_colormap(allow_random_colors=True)  # keyword has to be specified
plotting.view_img(mp.fetch(), cmap=cmap, symmetric_cmap=False).open_in_browser()
```

Uses the number of regions to determine the random seed for reproducibility.

### DIFF
--- a/siibra/volumes/parcellationmap.py
+++ b/siibra/volumes/parcellationmap.py
@@ -690,7 +690,7 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
             name=f"Custom colorization of {self}"
         )
 
-    def get_colormap(self, region_specs: Iterable = None):
+    def get_colormap(self, region_specs: Iterable = None, *, allow_random_colors: bool = False):
         """
         Generate a matplotlib colormap from known rgb values of label indices.
 
@@ -698,6 +698,7 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
         ----------
         region_specs: iterable(regions), optional
             Optional parameter to only color the desired regions.
+        allow_random_colors: bool , optional
 
         Returns
         -------
@@ -710,6 +711,10 @@ class Map(concept.AtlasConcept, configuration_folder="maps"):
                 "matplotlib not available. Please install matplotlib to create a matplotlib colormap."
             )
             raise e
+        if allow_random_colors:
+            seed = len(self.regions)
+            np.random.seed(seed)
+            logger.info(f"Random colors are allowed for regions without preconfgirued colors. Random seee: {seed}.")
 
         colors = {}
         if region_specs is not None:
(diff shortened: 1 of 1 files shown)


dickscheid (commented):
I would not return random colors if a colormap is absent. I think it is sufficient if siibra fails, but with a readable Exception which catches this case.  An alternative would be to return a commonly used map, so that at least the result is always the same.

AhmetNSimsek:
> I would not return random colors if a colormap is absent. I think it is sufficient if siibra fails, but with a readable Exception which catches this case.  An alternative would be to return a commonly used map, so that at least the result is always the same.

@dickscheid I disagree. When plotting, the user can use a default colormap themselves. This is an instance method which implicitly suggests there is a preconfigured colormap for this specific ParcellationMap.

Using a random seed, as the pr does, ensures reproducibility too. And we inform the user it is a non native colormap. Additionally, there are maps that are not fully covered. This PR also ensures a reproducible way to color them instead of setting to them the background.

xgui3783:
I think the flag should be `fill_uncolored` or `fill_unset` or `fill_undefined`, and defaults to `False` (which it already does)

This indicates the action, fill color map, and the target, areas without a color. 

This flag also implies that this action is **not** random, but deterministic. (though it may change from version to version of siibra-python)

AhmetNSimsek:
> @AhmetNSimsek I agree with you that a default colormap is not needed, but I also think that a random colormap is not useful either. I definitely prefer to raise a well defined exception if we don't have a parcellaion-specific colormap to provide.

This is already done, please see https://github.com/FZJ-INM1-BDA/siibra-python/commit/400520d261118e23ccff25a4791811b9be2986be.
However, this does not address the issue that if a map has only partial coverage, some regions have no color values but others do, the regions without colors will have the color of the background. I don't think this is correct.

Also, a reproducible random map is better than a standard colormap, IMO, since the odds that neighbor regions have similar colors are very slim. Not the mention, the user has to explicitly set it to True, in order to obtain such a colormap.

dickscheid:
> > @AhmetNSimsek I agree with you that a default colormap is not needed, but I also think that a random colormap is not useful either. I definitely prefer to raise a well defined exception if we don't have a parcellaion-specific colormap to provide.
> 
> This is already done, please see [400520d](https://github.com/FZJ-INM1-BDA/siibra-python/commit/400520d261118e23ccff25a4791811b9be2986be). However, this does not address the issue that if a map has only partial coverage, some regions have no color values but others do, the regions without colors will have the color of the background. I don't think this is correct.
> 
> Also, a reproducible random map is better than a standard colormap, IMO, since the odds that neighbor regions have similar colors are very slim. Not the mention, the user has to explicitly set it to True, in order to obtain such a colormap.

I was not aware that this should actually solve the missing region problem. Agreed then. Regarding optimal contrast of neighboring region colors, it is a complicated optimization problem so let's hope the random solution is sufficient.

PR #642, opened 2025-02-09
Rounding scheme for neuroglancer fetching needs revision

dickscheid:
When fetching an axis aligned flat volume of interest translating to exactly 1 voxel "thickness" in one dimension, the NeuroglancerScale fetching method rounds up to a 2 voxel array. In my example, fetching in bigbrain space with `Bounding box from (36.58,-0.10,8.33)mm to (38.64,-0.08,11.71)mm in BigBrain microscopic template (histology) space` retrieved an image array of shape (160, 2, 99).

This is critical, since several analysis workflows that I had implemented with siibra rely on the possibility to fetch a 2D patch from properly sized bounding boxes in physical coordinates, and would now confront me to resolve the ambiguity downstream.

The problem is caused by taking `floor`of the minpoint and `ceil` of the max point to compute the target shape as their difference, resulting in a doubling of integer rounding in the extreme case. The suggested change computes the rounded target shape as the `ceil` of the difference of the unrounded max- and midpoints instead, and adds it to the `floor`ed midpoint to obtain the rounded max point, therefore always only triggering a single rounding effect. Using the fix, the aforementioned bounding box yields an array of shape `(160, 1, 98)` after fetching, as expected.

The previous explicit additional clipping of voxel dimensions smaller than 1 seemed incorrect to me, and did not influence the result after the fix anyways, so I removed it.

The proposed change fixes the said problem, but it should be carefully thought through and tested on other examples. I suggest to define a range of standard and border cases of VOIs to double-check that fetching works as expected:
- perfect cube should result in perfect cube
- VOI dimensions equal or smaller to the volume's pixel size (resolution_mm) should result in flat dimensions (dim=1) of the flat array (that's the example reported above)
- VOIs larger than the volume dimensions should be clipped exactly to the max dimension of the volume
- ...

### DIFF
--- a/siibra/volumes/providers/neuroglancer.py
+++ b/siibra/volumes/providers/neuroglancer.py
@@ -554,6 +554,10 @@ class NeuroglancerScale:
         x0, y0, z0 = (np.array(bbox_.minpoint) - data_min).astype("int")
         xd, yd, zd = np.ceil(bbox_.maxpoint).astype(int) - np.floor(bbox_.minpoint).astype(int)
         offset = tuple(bbox_.minpoint)
+        if voi is not None:
+            logger.debug(
+                f"Input: {voi.minpoint.coordinate}, {voi.maxpoint.coordinate}.\nVoxel space: {bbox_.minpoint.coordinate}, {bbox_.maxpoint.coordinate}"
+            )
 
         # build the nifti image
         trans = np.identity(4)[[2, 1, 0, 3], :]  # zyx -> xyz
--- a/siibra/volumes/providers/neuroglancer.py
+++ b/siibra/volumes/providers/neuroglancer.py
@@ -529,7 +529,7 @@ class NeuroglancerScale:
         if voi is None:
             bbox_ = _boundingbox.BoundingBox((0, 0, 0), self.size, space=None)
         else:
-            bbox_ = voi.transform(np.linalg.inv(self.affine))
+            bbox_ = voi.transform(np.linalg.inv(self.affine), space=None)
 
         # extract minimum and maximum the chunk indices to be loaded
         gx0, gy0, gz0 = self._point_to_lower_chunk_idx(tuple(bbox_.minpoint))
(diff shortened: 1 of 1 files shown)


siibra/volumes/providers/neuroglancer.py
```diff
@@ -531,13 +531,6 @@ def fetch(self, voi: _boundingbox.BoundingBox = None, **kwargs):
...
             bbox_ = voi.transform(np.linalg.inv(self.affine))
 
-        for dim in range(3):
-            if bbox_.shape[dim] < 1:
-                logger.warning(
-                    f"Bounding box in voxel space will be enlarged to by {self.res_mm[dim]} along axis {dim}."
-                )
-                bbox_.maxpoint[dim] = bbox_.maxpoint[dim] + self.res_mm[dim]
-
         # extract minimum and maximum the chunk indices to be loaded
```
dickscheid, line 534: I think that the clipping was not correct  - if we know we are in voxel space here, and the intention is to ensure dim >= 1, we should set it right away rather than adding res_mm and relying on the rounding in the code below to do the rest. However, in the cases I considered, this block had no influence on the result anyways.

xgui3783:
> When fetching an axis aligned flat volume of interest translating to exactly 1 voxel "thickness" in one dimension, the NeuroglancerScale fetching method rounds up to a 2 voxel array. In my example, fetching in bigbrain space with Bounding box from (36.58,-0.10,8.33)mm to (38.64,-0.08,11.71)mm in BigBrain microscopic template (histology) space retrieved an image array of shape (160, 2, 99).

This is by design, no? We precisely do **not** want the user to deal in voxel space. So when bbox is provided, we try to satistify the user's request based on:

- user's request of resolution level
- the _smallest_ collection of voxels that _intersects_ with user's defined bbox

This will produce some weird edge cases - which is significantly magnified when user provides bbox with dimension similar or smaller than that of the voxel size. 

In your case, if you _definitely_ want a single voxel thickness, I would argue you would be fetching from (36.58,-0.09,8.33)mm to (38.64,-0.09,11.71)mm

> I suggest to define a range of standard and border cases of VOIs to double-check that fetching works as expected:

The two guidelines above aimed to do just that (but perhaps poorly documented)

> perfect cube should result in perfect cube

This is already like not possible, since the 1um slices cannot have perfect cube larger than 1 x 1 x 1

> VOI dimensions equal or smaller to the volume's pixel size (resolution_mm) should result in flat dimensions (dim=1) of the flat array (that's the example reported above)

This will create weird edge cases where we now need to make a decision, if the bbox sits **exactly** in the middle of two voxels. We will have to **at least** create one more rule to disambiguate this edge case

AhmetNSimsek:
(36.58,-0.10,8.33)mm to (38.64,-0.08,11.71)mm corrosponds to voxels bbox from (3330.127590551181, 3495.9708333333338, 5164.0913700787405) to (3170.4425511811023, 3494.9708333333338, 5066.7685354330715). While the difference for the y axis have exactly one voxel length, the question is defining how exactly siibra should handle data with 1 voxel length and intersecting with two voxels. Currently, siibra returns both of the voxels by taking the floor of  the lower and ceil of the higher. One can use round for both to return exactly one voxel but this standard is not defined in siibra yet. This PR will be morphed to determine this standard.

PR #652, opened 2025-03-09
Revise Figure 5 code example

dickscheid:
I rewrote the code example to make it closer to the manuscript figure and better understandable:

- except for the initial raw voxel plot, all plots are done in physical coordinates using nilearn.plotting, removing all explicit point to voxel conversions. I achieved this by finding out how we can work with the 2D canvases of nilearn's 3D plot to plot contours, points, bounding boxes
- The region of interest is not any more explicit, but defined just from a bounding box around a probability threshold of the pixels, so more intuitive now.
- complicated filtering and aggregation of layer 4 vertex projections now much simplified. filtering removed since the plotting command filters implicitly.
- overview plot with section position on glass brain now also implemented.

The PR should work, I tested several times, but please check it.

### DIFF
--- a/examples/tutorials/2025-paper-fig5.py
+++ b/examples/tutorials/2025-paper-fig5.py
@@ -46,8 +46,7 @@ assert siibra.__version__ >= "1.0.1"
 # First we retrieve the probability map of a motor area
 # from the Julich-Brain cytoarchitectonic atlas.
 region = siibra.get_region("julich 3.0.3", "4p right")
-probmaps = region.parcellation.get_map("mni152", "statistical")
-region_map = probmaps.get_volume(region)
+region_map = region.get_regional_map("mni152", "statistical")
 
 # %%
 # We can use the probability map as a query to extract 1 micron resolution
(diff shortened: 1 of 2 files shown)


siibra/features/image/sections.py
```diff
@@ -83,6 +83,14 @@ def __repr__(self):
     def section(self) -> CellbodyStainedSection:
         return self._section
 
+    def get_boundingbox(self):
```
AhmetNSimsek, line 86: fetch_kwargs must be passed

examples/tutorials/2025-paper-fig5.py
```diff
@@ -35,81 +35,126 @@
...
 assert siibra.__version__ >= "1.0.1"
 
 # %%
-# 1: Retrieve probability map of a motor area in Julich-Brain
+# First we retrieve the probability map of a motor area
+# from the Julich-Brain cytoarchitectonic atlas.
 region = siibra.get_region("julich 3.0.3", "4p right")
-region_map = siibra.get_map("julich 3.0.3", "mni152", "statistical").get_volume(region)
+probmaps = region.parcellation.get_map("mni152", "statistical")
+region_map = probmaps.get_volume(region)
```
AhmetNSimsek, line 50: why not
```python
region_map = region.get_regional_map("mni152", "statistical")  # this is a volume already
```

examples/tutorials/2025-paper-fig5.py
```diff
@@ -35,81 +35,126 @@
...
-# The section was intersected with the cortical layer 4 surface to get an approximation of
-# the mid cortex. This can be done by fetching the layer surface meshes, and intersecting
-# them with the 3D plane corresponding to the brain section.
+# Next we plot the section itself and identify the larger region of
+# interest around the patch, using the bounding box
+# of the centers of most relevant patches with a bit of padding.
+patch_locations = siibra.PointCloud(
+    [tuple(p.get_boundingbox().center) for p in patches],
+    space='bigbrain'
+)
```
AhmetNSimsek, line 90:
```python
patch_locations = siibra.PointCloud.union(*[p.get_boundingbox().center for p in patches])
```

AhmetNSimsek (changes requested):
Overall good but small changes could polish further. Please see my comments. (I can push these changes)

PR #656, opened 2025-04-02
Proper handling of VOI specifications when fetching across neuroglancer fragment volumes

dickscheid:
When fetching from a multi-fragment neuroglancer provider with a voi, the current implementation returned a full-size but almost empty volume. This PR tries to fix the problem, and had to address a couple of  downstream issues with neuroglancer fetching.

It needs testing and independent review...

The design of neuroglancer volumes, scales, providers etc. appears overcomplicated and makes it difficult to solve this elegantly, but solving the higher level design will be part of siibra-python v2.


siibra/volumes/providers/neuroglancer.py
```diff
@@ -205,35 +207,41 @@ def _merge_fragments(
...
                 result_arr = np.zeros(voxdims, dtype=img.dataobj.dtype)
                 result = nib.Nifti1Image(dataobj=result_arr, affine=result_affine)
 
             # resample to merge template and update it
-            resampled_img = resample_img_to_img(source_img=img, target_img=result)
+            try:
+                resampled_img = resample_img_to_img(source_img=img, target_img=result)
+            except BoundingBoxError:
+                logger.debug(f"Skipping fragment {frag_name} due to bounding box error")
+                continue
```
AhmetNSimsek, line 244: This might fail silently.

siibra/volumes/providers/neuroglancer.py
```diff
@@ -205,35 +207,41 @@ def _merge_fragments(
...
+        for frag_name, frag_vol in self._fragments.items():
             frag_scale = frag_vol._select_scale(
                 resolution_mm=resolution_mm,
                 bbox=voi,
                 max_bytes=kwargs.pop("maxbytes", NeuroglancerVolume.MAX_BYTES)
             )
             img = frag_scale.fetch(voi=voi)
+            if img is None:
+                logger.info(f"Fragment {frag_name} did not provide content for fetching.")
+                continue
```
AhmetNSimsek, line 227: I'd expect the fragment to supply a nifti with zeros

siibra/volumes/providers/neuroglancer.py
```diff
@@ -537,17 +543,25 @@ def fetch(self, voi: _boundingbox.BoundingBox = None, **kwargs):
...
+        data_zyx = None
         for gx in range(gx0, gx1):
             x0 = (gx - gx0) * self.chunk_sizes[0]
             for gy in range(gy0, gy1):
                 y0 = (gy - gy0) * self.chunk_sizes[1]
                 for gz in range(gz0, gz1):
+                    try:
+                        chunk = self._read_chunk(gx, gy, gz, kwargs.get("channel"))
+                    except DataAccessError:
+                        continue
```
AhmetNSimsek, line 555: Same as above

AhmetNSimsek (commented):
It is difficult to reproduce the issue. Could you provide a code? We can then write a test as well

IMO, this PR would cause silent failures.

siibra/volumes/providers/neuroglancer.py
AhmetNSimsek: This is ensured by lines 233-237

siibra/volumes/providers/neuroglancer.py
AhmetNSimsek: The debug message will be updated in the next commit

PR #685, opened 2025-11-04
Marmoset Interareal Connectivity Feature Support

baishi:
This feature enables retrograde tracer display of connectivity data between areas.


AhmetNSimsek (approved):
Given the constraints of siibra-python 0.4 and the specialized connectivity module in the current siibra-explorer, this solution relies on some non-ideal but required workarounds. But it will be adopted until siibra-python version 2 is released.

PR #702, opened 2026-04-30
maint: docs.yml

AhmetNSimsek:


.github/workflows/docs.yml
```diff
@@ -1,7 +1,7 @@
 name: '[Docs] Build docs and upload artifacts'
 on:
   push:
-    branches: [ 'main', 'doc_*', 'v1'] 
+    branches: [ 'main', 'doc_*', 'v1' ]
   release:
     types: [ 'published' ]
   workflow_dispatch:
```
xgui3783: I would rather the workflow not be trigger-able via API. this allows workflow to be run on any branch. Not an issue yet, but can be problem

.github/workflows/docs.yml
```diff
@@ -14,54 +14,67 @@ jobs:
...
       - name: Clear siibra cache
         continue-on-error: true
         run: |
           python -c "import siibra; siibra.cache.clear()"
-
       - name: Remove checked-out repository files
         continue-on-error: true
         run: |
-          rm -rf ${{ github.workspace }}/*
+          rm -rf ${{ github.workspace }}/*
```
xgui3783: also `rm -rf /tmp/*` please

xgui3783 (changes requested):
please see comments

PR #712, opened 2026-07-10
update doc and example workflows to use uv

AhmetNSimsek:


xgui3783 (approved):
I'm not familiar with uv action, so I don't know what kind of usage patterns it entails. 

I would advise monitor disk usage, and review/revise usage periodically.

PR #713, opened 2026-07-23
fix: use archive mode for siibra-configuration gitlab repo

xgui3783:
on siibra-api, we are hitting a lot of 429 otherwise.


xgui3783:
@AhmetNSimsek the PR is ready. Can you take a look?

Ideally we merge early tomorrow, and make a new release. 

at the moment, we cannot release siibra-api, because I run into 429. 

edit: if you find PR suitable, just approve it, and I can do the rest.

PR #724, opened 2026-08-28
Maint decoders

AhmetNSimsek:
This PR refactors retrieval decoders to support both file- and byte-based decoding.

Key changes:

* Introduces a `Decoder` abstraction with `FILE` and `BYTES` input modes.
* Allows decoders to define kwargs passed to the underlying reader.
* Adds explicit gzip handling:
  * native support when provided by the reader
  * cached decompression otherwise
* Preserves meaningful file suffixes in cache filenames.
* Moves path-capable formats such as NIfTI, GIFTI, FreeSurfer annotations, HDF5/NWB, NumPy, and tabular data to file-based decoding where appropriate.
* Removes unnecessary file → bytes → file roundtrips and enables file-backed/lazy access where supported.
* Keeps `find_suitable_decoder()` as a backwards-compatible wrapper around `Decoder.from_filename()`.
* Fixes `ZipfileConnector` to use `ZipfileRequest` directly instead of the bespoke `ZipFileLoader`, which was passing raw bytes to `FILE`-type decoders. `ZipfileRequest` correctly handles both decoder modes: `BYTES` decoders receive in-memory bytes, while `FILE` decoders have the zip member extracted to the siibra cache first and receive the path.
* Fixes `RepositoryConnector.get_loader` to pass the resolved `Decoder` directly to `HttpRequest` rather than wrapping it in a lambda, which also caused raw bytes to be passed to `FILE`-type decoders.
* Changes `ZipfileRequest` member matching from suffix-based to exact name matching, consistent with how `ZipfileConnector.search_files` returns full member paths.

Custom `func=` callbacks retain the existing byte-based behaviour.

### DIFF
--- a/siibra/configuration/configuration.py
+++ b/siibra/configuration/configuration.py
@@ -42,7 +42,7 @@ class Configuration:
         conn(
             server_or_owner,
             project_or_repo,
-            reftag="siibra-{}".format(__version__),
+            reftag="master",
             skip_branchtest=True
         )
         for conn, server_or_owner, project_or_repo in CONFIG_REPOS
--- a/siibra/retrieval/requests.py
+++ b/siibra/retrieval/requests.py
@@ -18,20 +18,22 @@ import json
 from zipfile import ZipFile
 import requests
 import os
+import shutil
+from pathlib import Path
 import gzip
-from io import BytesIO
 import urllib.parse
-from typing import List, Callable, TYPE_CHECKING
+from typing import List, Callable, TYPE_CHECKING, Literal, Optional
 from enum import Enum
 from functools import wraps
 from time import sleep
 import sys
+from dataclasses import dataclass, field
 
 from filelock import FileLock as Lock
 import numpy as np
 import pandas as pd
 from skimage import io as skimage_io
-from nibabel import Nifti1Image, GiftiImage, streamlines, freesurfer
+from nibabel import load as load_nibabel, streamlines, freesurfer
 import h5py
 
 from . import exceptions as _exceptions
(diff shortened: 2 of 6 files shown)


siibra/configuration/configuration.py
```diff
@@ -42,7 +42,7 @@ class Configuration:
         conn(
             server_or_owner,
             project_or_repo,
-            reftag="siibra-{}".format(__version__),
+            reftag="master",
```
AhmetNSimsek, line 45: will be reverted

siibra/retrieval/requests.py
```diff
@@ -53,81 +55,339 @@
...
-    url : str
-        The url or filename with extension.
-
-    Returns
-    -------
-    Callable or None
+    func : Callable
+        Function performing the actual decoding. Its first positional argument
+        receives either a filename or bytes according to ``input_type``.
+    input_type : {"FILE", "BYTES"}
```
xgui3783, line 98: does this mean input will either be "path to file" or "bytes"?

Also, preferred or enforced? From the desc func, it sounds to me that it is enforced/expected.

siibra/retrieval/requests.py
```diff
@@ -53,81 +55,339 @@
...
+        Preferred representation passed to ``func``. ``FILE`` allows readers
+        which support file-backed or memory-mapped access to operate directly
+        on the cached resource.
+    gzip : {"native", "decompress"} or None
```
xgui3783, line 102: I think this param is ambiguous. My first intuition is that this *file* is gzipped, (even though I am aware we are talking about decoders here)

Perhaps call it `can_gunzip` or `auto_gunzip` and boolean instead?

WRT to `decompress` -> cache etc, I think this is responsibility overload. It is not a Decoder's responsiblity to define how or if the source is cached.

siibra/retrieval/requests.py
```diff
@@ -53,81 +55,339 @@
...
+                filename = self._gunzip_to_cache(filename)
+            else:
+                with gzip.open(filename, "rb") as f:
+                    return self.func(f.read(), **self.kwargs)
+
+        if self.input_type == "FILE":
+            return self.func(filename, **self.kwargs)
+
+        with open(filename, "rb") as f:
+            return self.func(f.read(), **self.kwargs)
```
xgui3783, line 222: not issue per se, but this part of the code is rather difficult to parse what is is trying to do, and I would wager error prone, if we start to add more code and more branches to the dataclass

xgui3783 (approved):
If you have tested enough, then go for it. 

My concern is commented.

